PR #63, opened 2014-08-10
Mesh on lan

tcatm:


NeoRaider:
I don't like the reorganisation of the luci-portconfig page: when both IPv4 and IPv6 are set to 'static', there are no separators between the input fields, making the page a bit confusing.

PR #83, opened 2015-01-10
add section "system" to gluon-node-info with new option "role" and luci package to change role via config mode

kokel:
This Pull Request adds the section "system" to /etc/config/gluon-node-info and a new option "role" within "system".

Inside the mesh we want to distinguish normal nodes from other ones like backbone nodes, service nodes (servers like the one for the node map) and gateways.
This will give a facility to react differently on different nodes, especially at node map:
 - node detail view
 - node counts/statistics
 - display backbone links in another way
   - see http://map.freifunk-mwu.de/graph.html#mz-uniklinik for graph
   - see http://map.freifunk-mwu.de/geomap.html#lat=49.9926050006&lon=8.2622981071472 for geomap
 - display separate lists (see http://map.freifunk-mwu.de/list.html)

Maybe other things are possible at the backend while parsing alfred data in future.

A new section in site.conf will be necessary:

```
...
  roles = {
    default = 'node',
    list = {
        node = 'Normaler Knoten',
        backbone = 'Backbone Knoten',
        service = 'Service Knoten',
    },
  },
....
```
This is just an example and every community will be able to define here their own node roles as they like.

Node role:
Depends on the new section "site.roles" in the site.conf. In case the role is not set in `/etc/config/gluon-node-info` the invariant script "010-node-role" will set the role to "site.roles.default".

Package "gluon-luci-node-role":
Depends on gluon-node-info. Will add a new form to the expert config mode where it is possible to select the node roles specified in site.conf from a dropdown list. See the attached screenshot.
node_role

### DIFF
--- a/gluon/gluon-luci-node-role/Makefile
+++ b/gluon/gluon-luci-node-role/Makefile
@@ -1,0 +1,37 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gluon-luci-node-role
+PKG_VERSION:=0.1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(GLUONDIR)/include/package.mk
+
+define Package/gluon-luci-node-role
+  SECTION:=gluon
+  CATEGORY:=Gluon
+  DEPENDS:=+gluon-luci-admin +gluon-node-info
+  TITLE:=UI for specifying node role
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/gluon-luci-node-role/install
+	$(CP) ./files/* $(1)/
+endef
+
+define Package/gluon-luci-node-role/postinst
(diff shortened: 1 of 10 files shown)


tcatm:
I've reviewed the changes and I think it looks fine now. I'd only change the site's `node_roles` to `roles`.

Could you squash the commits a little (maybe one for the luci stuff and one for gluon-node-info changes)?

kokel:
Thank you for the review.
Two Points:
 - The text and title of the expert mode section is fine?
 - Is it possible to squash the commits in retrospect?
  - If yes, how can I do this?
  - Or is it quit easier to open a new PR for this?

tcatm:
You can squash commits using `git rebase -i` (http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) and force push to your branch. There's no need for a new PR.

I think the text is fine for now. We can tweak it a little after we've seen how this feature is used by communities.

kokel:
Thanks for the link.

The commits are now squashed into two clean commits:
 - One for the enhancements of "gluon-node-info"
 - One for the new package "gluon-luci-node-role"

I have renamed the site.conf section "node_roles" to "roles",  also.

PR #94, opened 2015-01-21
Simplified upgrade process

NeoRaider:
This rather large patchset reorganizes our upgrade infrastructure.

Support for initial scripts is dropped, the directory `/lib/gluon/version` isn't used anymore.

There is a new `sysconfig` variable, `gluon_version`, which is set at the end of the upgrade and thus contains the previous Gluon version during the upgrade. While the actual version number in the variable doesn't have any meaning, there are two special values that is checked for in some scripts:
* nil (unset): initial upgrade
* `legacy`: legacy upgrade

In addition, there is a script that sets the variable to the empty string when it is unset and `/lib/gluon/version/core` exists to differentiate this case from an initial upgrade.

All invariant and legacy scripts are moved to the common directory `/lib/gluon/upgrade` and are renumbered to be executed in the correct order. This allows to define the execution order of scripts across packages, an often requested feature.

All special handling of legacy upgrades is removed from `gluon-core`, legacy scripts are executed just like all other scripts (and check themselves if they have anything to do).

I'm posting this as a RFC as the legacy upgrade path hasn't been tested at all. Normal upgrades seem to work just fine. I'll also add some documentation when I've got some feedback.


NeoRaider:
I've now tested everything and made another pull request with some documentation.

In addition, I found out that the legacy upgrade is broken since b78daa7fb70919569acb4a1ae58ad7a878bb96d8, which I've also fixed while reworking scripts for this patchset...

NeoRaider:
Rebased onto master, including all new upgrade scripts.

PR #96, opened 2015-02-03
gluon-mesh-vpn-fastd: announce peer status using statistics.d

tcatm:
This adds basic peer statistics to statistics.d:

    "mesh_vpn": {
      "muehlentor": {
        "established": 23.8 // seconds
      },
      "huextertor": null,
      "holstentor": null
    }

This will be used to implement https://github.com/freifunk-gluon/gluon/issues/98

### DIFF
--- a/gluon/gluon-mesh-vpn-fastd/files/lib/gluon/announce/statistics.d/mesh_vpn
+++ b/gluon/gluon-mesh-vpn-fastd/files/lib/gluon/announce/statistics.d/mesh_vpn
@@ -1,0 +1,32 @@
+local nixio = require('nixio')
+local ltn12 = require('luci.ltn12')
+local json = require('luci.json')
+local uci = require('luci.model.uci').cursor()
+
+local socket_path = uci:get('fastd', 'mesh_vpn', 'status_socket')
+local fastd_sock = nixio.socket('unix', 'stream')
+
+if not fastd_sock:connect(socket_path) then
+  return nil
+end
+
+decoder = json.Decoder()
+ltn12.pump.all(ltn12.source.file(fastd_sock), decoder:sink())
+
+local status = decoder:get()
+
+local output = {}
+
+for key, peer in pairs(status.peers) do
+  local name, valid = peer.name:gsub('^mesh_vpn_backbone_peer_', '')
+  if valid == 1 then
+    if peer.connection then
+      output[name] = {}
+      output[name].established = peer.connection.established/1000
+    else
+      output[name] = json.null
+    end
+  end
+end
+
+return output


gluon/gluon-mesh-vpn-fastd/files/lib/gluon/announce/statistics.d/mesh_vpn
```diff
@@ -0,0 +1,28 @@
...
+
+decoder = json.Decoder()
+ltn12.pump.all(ltn12.source.file(fastd_sock), decoder:sink())
+
+local status = decoder:get()
+
+local output = {}
+
+for key, peer in pairs(status.peers) do
+  local name, valid = peer.name:gsub('mesh_vpn_backbone_peer_', '')
```
NeoRaider: The search expression should be anchored with `^`

gluon/gluon-mesh-vpn-fastd/files/lib/gluon/announce/statistics.d/mesh_vpn
```diff
@@ -0,0 +1,28 @@
...
+  if valid == 1 then
+    output[name] = {}
+    output[name].connected = peer.connection ~= null
```
NeoRaider: I think you mean `nil` here...

gluon/gluon-mesh-vpn-fastd/files/lib/gluon/announce/statistics.d/mesh_vpn
```diff
@@ -0,0 +1,28 @@
+local nixio = require('nixio')
+local ltn12 = require('luci.ltn12')
+local json = require('luci.json')
+local uci = require('luci.model.uci').cursor()
+
+local socket_path = uci:get('fastd', 'mesh_vpn', 'status_socket')
+local fastd_sock = nixio.socket('unix', 'stream')
+assert(fastd_sock:connect(socket_path))
```
NeoRaider: It would be nicer to just return `nil` when fastd is not running instead of causing an exception.

tcatm:
I've applied your suggestions.

NeoRaider:
Hmm, does it make sense to have both the fields `connected` and `established`?

tcatm:
Both fields are present when `connected` is `true`. When it's `false` only `connected` is returned.

NeoRaider:
Yes, so `connected` is redundant and can be removed.

tcatm:
So it would be either `[Object]` (connected) or `undefined` (disconnected)?

NeoRaider:
Yes, something like that.

tcatm:
I don't think it's possible to create such JSON using lua. Setting `output[peername] = nil` will omit the peer.

NeoRaider:
Hmm, then just always use an object, which either has an integer "connected" or is empty?

tcatm:
I prefer the current solution as it can easily be extended to include additional information (e.g. traffic statistics) in the future.

NeoRaider:
That would work the same way with my suggestion. I'd just drop the value `connected` and rename `established` to `connected`, but of course it's still possible to add other values in the future.

tcatm:
Hrm. Checking whether an object is empty is rather clumsy in JavaScript (`Object.getOwnPropertyNames(o).length == 0`), however if we decide to do it this way I'd rather have the key named `established` instead of `connected`.

NeoRaider:
Leaving the name as `established` is okay.

Why do you want to explicitly check if the object is empty? Just check if `established` is undefined...

tcatm:
I've figured out how to create an explicit `null` value in lua.

PR #97, opened 2015-02-07
Adding WiFi settings to config mode making the private-wifi-page redunda...

CodeFetch:
Had no nerves to test compilation with gluon, but it should work unless I have made an extremely stupid mistake...
Integrates #93.


tcatm:
Please squash your commits. "Fixed bad ideas" is a rather meaningless commit message.

A longer explanation in the commit message would also help in understand the changes you are proposing making review much easier.

PR #98, opened 2015-02-07
Adding WiFi settings to config mode making the private-wifi-page redunda...

CodeFetch:
This commit removes the "Private WLAN"-settings of the configmode. 
Instead it adds a new "WLAN"-settings page which integrates the "Private WLAN"-settings and the proposed "wifi-config" #93.
Furthermore one can change the txpower, the distance setting, the SSID of the client network, select the region and whether the region should be exposed in the beacons' IEs.

As I already said I have had no nerves to test compilation with gluon, yet, but it should work unless I have made an extremely stupid mistake.


tcatm:
Thanks for working on this. I understand testing luci modules in Gluon is tedious due to the many reboots and upgrades required. This gets even more complicated due to the fact that you are the only one capable of pushing to your branch (CodeFetch:master). This makes testing and possibly fixing things for other developers even more cumbersome.

Nevertheless I did some static analysis on your code and I do have some remarks:

- private wifi and mesh wifi are tightly coupled. I'd rather keep it in seperate packages as private wifi will be expanded in future versions to support things like DHCP and routing.
- hwmode will never be `11na` or `11ng`. Whether it's `n` is determined by `htmode` (see http://wiki.openwrt.org/doc/uci/wireless)
- setting the default value of `country_ie` should be moved to package `gluon-mesh-batman-adv-core`, complete with a `site.conf` option. This should be a separate pull requests.
- the f.handle function has too many comments. I think the `uci:set(...)` statements are pretty clear on their own.
- Consider using a luci Map (see http://luci.subsignal.org/trac/wiki/Documentation/CBI) and Compound. A Map will take care of reading and writing config options so you'll have to provide labels.

CodeFetch:
Okay, I'll revise it.

I know about the planned features for the private wifi, but as far as I understood it shouldn't become much more than a simple static NAT. Personally I dislike the idea of having a separate form for that.

I agree that there should be a site.conf option for country_ie.
Easy access to the country_ie-setting in configmode is also important as it affects clients. 
If one changes the region setting to get a higher tx output to e.g. compensate cable loss and country_ie is enabled then a client device might think it's in the USA or something and exceeds the maximum output power concerning country regulations. 
This can also have wacky side effects like low battery life.

First I used a luci map, but it didn't work properly. After having saved changes, the form still showed up with the old settings and I needed to refresh the page to see the actual changes. I gave up on that, but I  can give it a second try.

do9xe:
What will happen here in the future? looks like it makes my work on #93 useless.

CodeFetch:
That's a good question. Looks like tcatm doesn't want a Q&D solution and to be honest me neither.
In my opinion we should completely rework the network and wifi settings in expert mode.
We just need a network-tab for DHCP, NAT and Bridging that recognises and prevents conflicts and one tab for wifi settings and private wlan. Then we can just add a checkbox for private wifi to enable DHCP and NAT and change settings under the network-tab if desired. That's my suggestion. 
This would do what https://github.com/freifunk-gluon/gluon/issues/284 suggests.

Just make a pull request to my fork and I'll merge changes if you want to.

do9xe:
I would suggest it the other way around.
We should definitly seperate the privat-wifi and the wifi-config settings into two different packages.
I know, that @NeoRaider wanted to add the tx-power settings to wifi-config. So let's leave it as it is.

for the private-wifi I would suggest to add all this into the private wifi package, as it is an addon, which comes with both good and bad things. So i suggest for the privat-wifi we could rewrite it to this:
* checkbox for en/disable
* Fields for SSID & Key
* dropdown for NAT/bridged
* checkbox for the DHCP
* textedit for DHCP settings

We can talk in the IRC about it, if you want. :) im in "gluon" and "ffki" on hackint.

CodeFetch:
We need a separate section for networking anyway. 
Currently we have an option for Mesh-on-WAN, but as NeoRaider suggests in https://github.com/freifunk-gluon/gluon/issues/284 we should have different options for bridging. Furthermore we should be able to set port forwarding rules.
Another thing is that many people complain gluon isn't actually usuable for the purpose it was originally intended to be used for. Without a queen there is currently no DHCP-server. 
Which is really stupid as the IPv6-addresses are generated from the MACs by the server and are thus predictable. We need a function to generate IPv6-addresses by MAC for the specific community that is configurable in the site.conf and a fallback DHCP-server.
With all desired features implemented we'd have so many options that we should categorise them and to me differentiating between WiFi and Networking options makes totally sense.

I'm just in my test phase and can't work on it right now, but I'll take care of it in about two weeks.

NeoRaider:
@CodeFetch, IPv6 addresses are never generated by a server, each client generates its addresses itself. The network doesn't have any control over the way the clients generate their addresses in the stateless autoconfiguration model.

By design, each client should have at least a static address generated from its MAC, but may use any number of additional addresses. Most current operating systems use the so-called IPv6 privacy extensions, which use randomly generated IP addresses which are changed at regular intervals. The random IP addresses are used for all outgoing connections, so if you don't tell anybody about your static IPv6 address, nobody can find out about it.

If your clients don't use random IPv6 addressses, you should find out how to enable the privacy extensions. I know that Windows and most Linux distributions have them enabled by default, not sure about MacOS and mobile devices...

CodeFetch:
Okay... Seems that I haven't fully understood IPv6 especially concerning stateful and stateless address configuration, yet.
So the DHCPv6 server only gives the prefix for the address the client generates either randomly or using the MAC address?! Would it be easily possible to configure a DHCP server in OpenWRT to return the correct prefix for the community to allow clients to get a DHCP lease that can still be used when the node's connection to the gateway servers is up again?
I mentioned this as a problem because at least Windows and Gnome Network Manager users need to explicitly set a static IP-address when a gluon node has no connection to a DHCP server. 
By default their systems retry to establish a connection until they communicated with a DHCP server.

tcatm:
There's no DHCPv6 involved. It's just Router Advertisements and those are already distributed by the nodes itself.

As far as IPv4 is concerned there are plans to create a distributed DHCP server to run on the nodes themself but at this time it's unclear when that will happen and whether IPv4 will be fully deprecated first.

CodeFetch:
Sorry, it's off-topic again.
In my community (Freifunk Rheinland) we have a DHCPv6 server.
As far as I'm concerned we need it because our IPv6 range isn't NATed.

do9xe:
@CodeFetch what about your PR?
Do you still want it to get merged? You got totally OffTopic.
It seems like you want a seperated networking section, rather than the package mixup you made with this PR.

NeoRaider:
Closing for now, we're changing the repo structure.

PR #100, opened 2015-02-09
Added a 120 second timeout for wget to prevent it from stalling forever.

eriktews:
Hi, I sometimes see a lot of stalling wget processes from my autoupdater on my freifunk node. Specifying -T 120 should make wget terminate after 120 seconds. 

When the busybox in gluon has not been built with the timeout option, wget seems to ignore the -T 120 option.

This patch is expected to work fine, but is not yet tested.

### DIFF
--- a/admin/autoupdater/files/usr/sbin/autoupdater
+++ b/admin/autoupdater/files/usr/sbin/autoupdater
@@ -125,7 +125,7 @@ local function read_manifest(mirror)
 
   -- Read all lines from the manifest
   -- The upper part is saves to lines, the lower part to sigs
-  for line in io.popen(string.format("wget -O- '%s/%s.manifest'", mirror, branch.name), 'r'):lines() do
+  for line in io.popen(string.format("wget -T 120 -O- '%s/%s.manifest'", mirror, branch.name), 'r'):lines() do
     if not sep then
       if line == '---' then
 	sep = true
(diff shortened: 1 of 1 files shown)


tcatm:
The commit message should include the package name (e.g. autoupdater: add 120s timeout to wget). What's the default? I can't imagine wget would retry forever.

eriktews:
A normal wget would timeout after a while. However, the busybox wget doesn't do that. When it is not compiled with the timeout option or the -T parameter is not specified, it tries forever:

http://git.busybox.net/busybox/tree/networking/wget.c

Also, when I logged into a freifunk node that was running for 2 days, I saw about 5 autoupdater processes running, all waiting for the termination of wget. So I assume that problem might also cause some nodes to crash after a long uptime, when too many wget processes are running on the system.

I also created a pull request for gluon itself, so that wget there is compiled with the timeout option:

https://github.com/freifunk-gluon/gluon/pull/292

The commit message has been updated.

tcatm:
I see. Thanks for looking into this. I think the dependency on `CONFIG_BUSYBOX_CONFIG_FEATURE_WGET_TIMEOUT` should be moved to the autoupdater's Makefile itself so it'll work correctly outside of gluon.

NeoRaider:
I don't think it makes sense to have that as a hard dependency, you should also be able to use the autoupdater with a completely different wget implementation (like the original one).

tcatm:
That's a good point. Yet, if the supplied wget does not accept `-T` it might render some remote nodes unaccessible.

eriktews:
Hi, it looks to me like, that wget doesn't support -T in busybox, the -T option will just be ignored. So it should be fine when only one of the two pull requests are merged.

eriktews:
About the dependency, autoupdater currently doesn't even depend on wget at all. So I would suggest not to make this a hard dependency.

eriktews:
So the change to the gluon master has been mergen:

https://github.com/freifunk-gluon/gluon/pull/292#event-232780334

At least from now on, wget in gluon master will have a timeout option.

tcatm:
It looks like busybox wget will indeed ignore `-T` so this should be safe to merge.

eriktews:
I have also created a separate branch with a dependency on the WGET_TIMEOUT option. Feel free to merge also this branch:

https://github.com/eriktews/packages/tree/wget-timeout-dependency

PR #102, opened 2015-03-01
add legacy-support for Internet Explorer

dracoTrier:
There are still a bunch of Windows Phone or Internet Explorer users out
there. This shows an [IE not supported, click here for legacy......]
link and adds the old status page for these users. (can also be used by
users without IPv6-support (eg. many Android wlan-users).


tcatm:
Have you considered Server-Sent Events polyfills? Support for the most recent IE version would be sufficient.

PR #103, opened 2015-03-02
provide more info about the node performing an autoupdate

cbricart:
this patch is adding a $QUERY-string to the request for the manifest file, providing some more information to the update server about the node that wants to autoupdate.
This allows the (web-)server to switch between multiple manifests according to criteria like a MAC-list.

Possible scenarios include to automatically pull a node into another domain via it's autoupdate based on it's known MAC address the server switches for

### DIFF
--- a/admin/autoupdater/files/usr/sbin/autoupdater
+++ b/admin/autoupdater/files/usr/sbin/autoupdater
@@ -7,6 +7,11 @@ local platform_info = require('platform_info')
 local uci = require('luci.model.uci').cursor()
 local util = require('luci.util')
 
+-- for site.site_code
+local site = require 'gluon.site_config'
+-- for sysconfig.primary_mac
+local sysconfig = require 'gluon.sysconfig'
+
 local autoupdater_util = require('autoupdater.util')
 local autoupdater_version = require('autoupdater.version')
 
@@ -123,9 +128,14 @@ local function read_manifest(mirror)
 
   local ret = {}
 
+  -- more info about the note to the update server:
+  local site = site.site_code
+  local primary_mac = sysconfig.primary_mac
+  local meshvpn_enabled = uci:get("fastd", "mesh_vpn", "enabled", "0")
+
   -- Read all lines from the manifest
   -- The upper part is saves to lines, the lower part to sigs
-  for line in io.popen(string.format("wget -T 120 -O- '%s/%s.manifest'", mirror, branch.name), 'r'):lines() do
+  for line in io.popen(string.format("wget -T 120 -O- '%s/%s.manifest?mac=%s&mesh=%s&site=%s'", mirror, branch.name, primary_mac, meshvpn_enabled, site), 'r'):lines() do
     if not sep then
       if line == '---' then
 	sep = true


NeoRaider:
The autoupdater must have no dependencies on Gluon packages as we plan to move it upstream some time in the future. (This holds for all packages outside the `gluon` subdir of the package repo)

Adding a generic `query_string` field to `/etc/config/autoupdater` which is set by a script in the package `gluon-autoupdater` would allow adding a query string without this dependency.

cbricart:
agreed - I'll rework the patch

just for additional info: any direct namespace which would be also prohibited via `uci:get()` ?

mruettgers:
The value of mesh_vpn is no reliable indicator for the current online state of a node.
It may be set to 1 even without any wan uplink.

NeoRaider:
I think the query string should be static (I don't see a sane way to change it when the VPN is enabled or disabled anyways), so just the node ID and the site code would be enough.

The node ID should be used in favour of the primary MAC address in my opinion (for now they are always equal, but that might change in the future when we support more unusual hardware).

mruettgers:
I suggest to add the status of the WAN interface, to determine if a node is able to fetch an update itself.
```
local state = require('uci').cursor(nil, "/var/state")
local wan_up = state:get("network", "wan", "up", "0")
```
The interface could probably set in a configuration option.

NeoRaider:
@mruettgers, as I wrote before, the autoupdater should not depend on such state. The query string should be entirely static and determined directly after flashing.

tcatm:
@mruettgers Why would the WAN status be of any interest anyway? We're building a mesh based network where nodes are encouraged to use wireless links to fetch updates. That's a primary feature of Freifunk networks.

The WAN link is not used for fetching the update directly.

cbricart:
@tcatm actually this is not true - nodes with WAN-uplink *do* connect to the update servers using their WAN-interface/IP. If this is unintentional then this is a bug!

second, why should the info about WAN status of a node matter? well, if the update will include a change for the wireless settings (i.e. channel, ESSID, ..) the update server has to have the info to be able to update/migrate meshing-only satellite nodes first because otherwise they might lose their only uplink when updating the wired nodes first

mruettgers:
Currently we have the situation, that we will transfer nodes from one community to another.
Due the fact, that we also change the WiFi-Channel and the Mesh-SSID, we have to insure that the Sattelites without an Uplink are updated first, because they loose their connection if the Uplink-Node has been updated before.

cbricart:
@NeoRaider so as outlined above, some things need to be dynamic and independent from the state the node had when flashing. currently I'm thinking about a cron-job installed by `gluon-autoupdater` that periodically sets some property using `uci set ..` - as long as there is no following `uci commit [..]` it does not mess with the flash. the information does not have to be persistent anyway as it is periodically refreshed

would that be an adequate approach?

tcatm:
@cbricart Looking at the site.conf in your ffac-site repo I'm seeing static IPv4 used in autoupdater URLs. This is not supported. IPv6 is required as nodes do not have any guaranteed IPv4 connectivity (except when connected via WAN).

A `uci set` without a commit might break when some other part of the system issues a commit.

@mruettgers The autoupdater can not handle such scenarios at this time. Our best idea is for nodes to connect as clients to neighbouring nodes in case they lose connectivity. There's no code do that yet, though.

NeoRaider:
Closing for now, we are changing the repository structure. Please open a ticket against the main repository if you want to continue this discussion.

PR #104, opened 2015-03-10
mesh-ba-core: remove dead code

tcatm:
hwmodes 11ng and 11na are invalid values:

http://wiki.openwrt.org/doc/uci/wireless


NeoRaider:
This isn't dead code, it is required for systems that have been upgraded from Attitude Adjustment.

PR #105, opened 2015-03-10
Split wireless config

tcatm:
This will be required for adding 11s support.


NeoRaider:
The second patch looks fine, but we should probably introduce some helper functions to reduce code duplication.

PR #106, opened 2015-03-16
ebtables-filter-mcast: drop ARPs to/from 0.0.0.0

tcatm:
Fixes https://github.com/freifunk-gluon/gluon/issues/311


NeoRaider:
This will also filter out legitimate ARP requests with 0.0.0.0 source (for example used during DHCP before an address has been assigned).

Only requests with 0.0.0.0 destination should be filtered (And maybe replies with 0.0.0.0 source? Are there any multicast ARP replies? Gratuitious ARP?)

tcatm:
Hrm, yes, I've missed matching the arp-opcode. Commit updated.

NeoRaider:
Looks fine now.

PR #111, opened 2015-03-24
Text to choose password much more insistent

rubo77:
add a small hint, to create a minimum 12 char long password with letters and numbers.


tcatm:
I don't think we need to make this any clearer. 24 characters are much longer than most people can remember.

rubo77:
I meant "mit mindestens 12 Buchstaben und/oder Zahlen"

mweinelt:
The 12 character recommendation sounds very arbitrary.

Other than that: The option is already hidden in the expert mode, and people using ssh probably know, at least to some degree, what they're doing.

PR #112, opened 2015-03-24
Enhancements in language file for mesh settings

rubo77:


tcatm:
What's the reason for adding the "if" in a place where it will never be visible for users? I'd suggest

rubo77:
Ah I see, it is only the id, I thought it would be visible to a potential english build, but seems I was mistaken. I just corrected it when I saw it. the main thing was the word "beschränken", so can this be merged? or do I have to remove the "if"?

NeoRaider:
As you changed the id, the translation won't match at all like this. 

Wo ist das Problem mit "beschränken"?

PR #116, opened 2015-04-13
first attempt to build in support PPPoE + local LAN port option

muebau:
Hi, I would like to add support for PPPoE as its sometimes needed it the router is used in a standalone manner.
The "local LAN" option is needed to it the user would like to use the FreiFunk router as the main device to provide the FreiFunk WiFi (+mesh) but would like to leave the devices at the LAN port as they are (just connected to the LAN, not FreiFunk).

I would need a hint where to put a NAT configuration as its needed for the PPPoE scenario.


NeoRaider:
You are mixing up two indepentent problems:
* Replacing the simple WAN uplink with a PPPoE-based uplink
* Providing a NATted LAN

I'd prefer to have a solid solution for freifunk-gluon/gluon#284 before we add new network configuration options. We might want to realize special configurations like PPPoE as independent Gluon packages, unless the configuration changes involved are trivial.

NeoRaider:
Closing for now as we are changing the repository structure.

PR #117, opened 2015-04-16
change network to 192.168.0.1/24

RubenKelevra:
on tp-link hardware, this is the default-network, on a flash of factory the original-firmware expect the new firmware with the same ip, which works nicely here.


kaechele:
Or in other words: After flashing a TP-Link Router from the original firmware to gluon (using a factory image) it will transition nicely into config mode after the router rebooted with no changes to the network config needed (just like when updating the factory firmware).

NeoRaider:
Nack. This isn't even the same on all TP-Link models, and there is a lot of other which use other addresses.

192.168.1.1 is at least the same as is used for the failsafe mode. Changing the setup mode address will cause so much confusion that we'd really need a very good reason to ever do that.

rubo77:
This is the better idea: https://github.com/freifunk-gluon/gluon/issues/339

PR #118, opened 2015-08-30
wifi-mesh-macfilter: Setting plink_action, when 802.11s mesh-interface comes up

kb-light:


kb-light:
x86-kvm_guest build failed with this package
ar71xx-generic is working

kb-light:
Does not work as desired.
If a connection is lost, the plink_action seems to be deleted, too.

rotanid:
did you have any idea how to solve this in the meantime?
first thought: cronjob.
but there may be a better way like some trigger...

mweinelt:
The plink_action is stored in the station entry¹, so if the station goes missing in its entirety so does the plink_action. For this package to work we'd currently need a userspace application managing the blocklist and tracking station events.

[1] https://elixir.bootlin.com/linux/v5.0-rc3/source/include/net/cfg80211.h#L981

NeoRaider:
I'd like to see support for this in hostapd...

maurerle:
An approach which does this using a cronjob and a user-defined list of ignored nodes can be found here:
https://github.com/Freifunk-Rhein-Sieg/freifunk-packages/blob/8e402bc4879e98ac7fe47735e60c34cd2d6c1a14/gluon-rsk-block-mesh/files/lib/rsk/blockmesh/blockmesh.sh

neocturne:
Has anyone checked if the hostapd config options `macaddr_acl` and `accept_mac_file`/`deny_mac_file` work for 11s?

neocturne:
> Has anyone checked if the hostapd config options `macaddr_acl` and `accept_mac_file`/`deny_mac_file` work for 11s?

wifi-iface UCI option `macfilter` (values `allow` or `deny`) + list `maclist` should set these options.

PR #120, opened 2015-09-04
simple-uci: wrapper around uci providing useful functions

tcatm:
- delete_all
- section
- tset
- get_bool
- get_list
- get_first
- add_to_set
- remove_from_set

### DIFF
--- a/libs/lua-simple-uci/Makefile
+++ b/libs/lua-simple-uci/Makefile
@@ -1,0 +1,33 @@
+#
+# Copyright (C) 2015 Nils Schneider <nils@nilsschneider.net>
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua-simple-uci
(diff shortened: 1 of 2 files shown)


libs/lua-simple-uci/Makefile
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright (C) 2015 Nils Schneider <nils@nilsschneider.net>
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=simple-json
```
NeoRaider: The PKG_NAME looks wrong...

NeoRaider: Also, the directory is called lua-simple-uci, but everywhere else this is just called simple-uci... I prefer the version with prefix or suffix, so it is clear that it is a Lua library.

NeoRaider:
Superseded.

PR #122, opened 2015-12-13
Add new packages sse-multiplex and batman-adv-visdata

NeoRaider:


NeoRaider:
Not tested of OpenWrt yet, but does compile now...

NeoRaider:
I'm currently debugging a weird crash in sse-multiplexd, so this is not ready yet.

NeoRaider:
The crash is fixed now.

NeoRaider:
I've decided to reimplement batman-adv-visdata in C before merging this.

rubo77:
Can you add a small description why this is needed?

Do server side events not work correctly in Freifunk?

NeoRaider:
@rubo77, see the description of the sse-multiplex package: https://github.com/freifunk-gluon/packages/blob/sse-multiplex/net/sse-multiplex/Makefile#L17

The idea is to allow multiple viewers of the status page, while only needing a single backend process generating the event stream (instead of one per viewer).

PR #123, opened 2015-12-21
autoupdater: Add posibility to specify mirrors on commandline

jplitza:
The list from the commandline isn't shuffled, so if mirrors are determined externally, they can be prioritized.

A possible application is detailed in freifunk-gluon/gluon#454. Also a manual update with a custom mirror, for example on a node that was offline for a long time, is now easily realizable.

### DIFF
--- a/admin/autoupdater/files/usr/sbin/autoupdater
+++ b/admin/autoupdater/files/usr/sbin/autoupdater
@@ -35,6 +35,7 @@ local force = false
 -- timespan given by the priority and another 24h have passed
 local fallback = false
 
+local mirrors = {}
 
 local function parse_args()
   local i = 1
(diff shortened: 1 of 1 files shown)


NeoRaider:
LGTM, do you think this should be merged even though we plan to replace the autoupdater?

jplitza:
Thanks for getting back to me on this one! ☺

Yes, for two reasons:
1. My last impression regarding the autoupdater-rewrite was, that it's stuck with SHA256 vs. SHA512 issues in the manifest/signing in particular, and won't happen anytime soon (as in, weeks/until 2017.1) in general.
2. Having this merged somehow sets the standard for the rewritten autoupdater to also support that feature, or at least is a commitment to have it in general (even if not in the first version of the rewrite).

PR #125, opened 2015-12-26
respondd: use C modules instead of Lua to provide data

NeoRaider:
This updates/rewrites respondd, dropping Lua in favour of simple C modules.

The data providers haven't been converted yet.


tcatm:
This needs documentation or at least a descriptive commit message so people know how to write providers for this new framework.

NeoRaider:
Yes, I've planned to write a README file.

Converting the data providers is almost finished. Some providers can't be converted nicely as we need site.conf data, so I'm currently considering converting site.conf to JSON during build...

The RAM usage hasn't improved as much as I had hoped (virtual memory usage has reduced to 2MB on ar71xx), but I'll try to further optimize that. The new C providers are extremely fast (so fast that even without the cache the data is returned without noticable delay).

tcatm:
I like the idea of converting `site.conf` to json during build. In general, it's probably a good idea to not use much scripting on nodes to avoid spikes in memory usage.

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+
+static void update_time(void) {
+	struct timespec tp;
+	clock_gettime(CLOCK_MONOTONIC, &tp);
+
+	now = (int64_t)tp.tv_sec * 1000 + tp.tv_nsec / 1000000;
+}
+
+
+/** Merges two JSON objects (on conflicts, object a will be preferred) */
```
jplitza: It might be noteworthy that this happens in-place in the second object.

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+
+		json_object_object_add(b, key, merge_json(val_a, val_b));
+	}
+
+	json_object_put(a);
+	return b;
+}
+
+static const struct respondd_provider_info * get_providers(const char *filename) {
+	char path[2 + strlen(filename) + 1];
```
jplitza: Shouldn't this be `strnlen(filename, NAME_MAX)`? Just to be safe?

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+	snprintf(path, sizeof(path), "./%s", filename);
```
jplitza: Probably everybody apart from me knows that, but I had to look up why `./` is prefixed. Maybe add a comment on that?

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
-  puts("  -g <ip6>         multicast group, e.g. ff02::2:1001");
-  puts("  -p <int>         port number to listen on");
-  puts("  -i <string>      interface on which the group is joined");
-  puts("  -c <string>      Lua command");
-  puts("  -h               this help\n");
+	puts("Usage: respondd [-h] -g <group> -p <port> -i <if0> [-i <if1> ..]");
+	puts("	-g <ip6>         multicast group, e.g. ff02::2:1001");
+	puts("	-p <int>         port number to listen on");
+	puts("	-i <string>      interface on which the group is joined");
+	puts("	-d <string>      data provider directory");
```
jplitza: It might be noteworthy that the cwd is taken as default if `-d` is not specified.

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+
+
 static void usage() {
-  puts("Usage: respondd [-h] -g <group> -p <port> -i <if0> [-i <if1> ..]");
-  puts("  -g <ip6>         multicast group, e.g. ff02::2:1001");
-  puts("  -p <int>         port number to listen on");
-  puts("  -i <string>      interface on which the group is joined");
-  puts("  -c <string>      Lua command");
-  puts("  -h               this help\n");
+	puts("Usage: respondd [-h] -g <group> -p <port> -i <if0> [-i <if1> ..]");
```
jplitza: This line doesn't follow my understanding of help messages at all:
* `-h` is actually effectively mutually exclusive with all other options.
* This message suggests that `-g`, `-p` and `-i` are all mandatory, but none of them are.

I'd suggest the following:
```
Usage:
  respondd -h
  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir>]
```

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+		load_cache_time(r, provider->request);
+
+		key.data = r;
+		entry = hsearch(key, ENTER);
+	}
+
+	struct request_type *r = entry->data;
+
+	struct provider_list *pentry = malloc(sizeof(*pentry));
+	pentry->name = strdup(name);
```
jplitza: Again, `strndup(name, NAME_MAX)`?

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+
+	pentry->next = *pos;
+	*pos = pentry;
+}
+
+static void load_providers(void) {
+	update_time();
+
+	/* Maximum number of request types, might be made configurable in the future */
+	if (!hcreate(32)) {
```
jplitza: I feel bad about using the non-reentrant versions of hash tables. What if some provider wants to use a hash table as well?

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
-  }
+	struct request_type *r = entry->data;
 
-  clock_gettime(CLOCK_MONOTONIC, &tp);
+	if (r->cache_time && now < r->cache_timeout)
+		return json_object_get(r->cache);
 
-  lua_pushlstring(L, buffer, read_bytes);
-  lua_pushnumber(L, tp.tv_sec + 1.0e-9*tp.tv_nsec);
+	fputs("Generating\n", stderr);
```
jplitza: Um… hu? This is debug output and should be removed, right?

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+
+	char filename[strlen(name) + 7];
+	snprintf(filename, sizeof(filename), "%s.cache", name);
+
+	FILE *f = fopen(filename, "r");
+	if (!f)
+		return;
+
+	float val = 0;
+	fscanf(f, "%f", &val);
```
jplitza: Why on earth would we use floats as cache periods? Isn't second-resolution high enough? Or are you planning ahead for caching periods longer than 100 years?
I would prefer `%u` here.

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+	socklen_t addrlen = sizeof(addr);
+	bool compress;
+
+	input_bytes = recvfrom(sock, input, sizeof(input)-1, 0, (struct sockaddr *)&addr, &addrlen);
+
+	if (input_bytes < 0) {
+		perror("recvfrom failed");
+		exit(EXIT_FAILURE);
+	}
+
```
jplitza: We could short-circuit here if `input_bytes == 0`

tcatm: Should we decide to do that, I'd like to see that implemented in `handle_request` instead.

net/respondd/src/respondd.c
```diff
@@ -41,150 +48,381 @@
...
+			output_bytes = compressed_bytes;
+		}
+	}
+	else {
+		output = str;
+		output_bytes = strlen(str);
+	}
+
+	if (output) {
+		if (sendto(sock, output, output_bytes, 0, (struct sockaddr *)&addr, addrlen) < 0)
```
jplitza: Isn't `< output_bytes` more honest? As we do not draw any consequences from the failed call, we might just as well check if the whole message was sent.

jplitza:
Great work, thanks! I added some comments and played around a bit locally with a minimal example, seems to work fine.
Apart from my code comments (I would rather use the length-aware string functions everywhere, even though probably not necessary), the readme you promised is still missing. ;-)

net/respondd/src/respondd.c
NeoRaider: Ok.

net/respondd/src/respondd.c
NeoRaider: NAME_MAX is an arbitrary limit, using it should be avoided. Programs ought to work with longer names.

net/respondd/src/respondd.c
NeoRaider: Ok.

net/respondd/src/respondd.c
NeoRaider: Ok.

net/respondd/src/respondd.c
NeoRaider: Makes sense. I didn't really change much about the help output (I've just updated the indentation), but this should be fixed.

net/respondd/src/respondd.c
NeoRaider: Oops, yes.

net/respondd/src/respondd.c
NeoRaider: If providers would want to use the hash tables, exactly one would be able to use the non-reentrant version of create... which should it be?

Choosing which code part of or loaded by respondd should be allowed to use hcreate, respondd would be the only sensible choice. I don't think there's a problem with this.

I might change this anyways, as memory managed by hcreate is annoyingly not seen by valgrind, so valgrind always reports lost allocations...

net/respondd/src/respondd.c
NeoRaider: I'd like to change statistics and neighbours cache times to something like 0.9 seconds, so a consumer updating the data once a second will always get new data. The new respondd is to much faster than the Lua version that sub-second cache times make actually sense.

We could change this to milliseconds, so no floats are needed. The Lua version used seconds as well, but that hasn't been in any Gluon release yet, so I don't see an issue with changing it.

net/respondd/src/respondd.c
NeoRaider: Doing this in `handle_request` seems fine to me.

net/respondd/src/respondd.c
NeoRaider: As sock is a datagram socket, sends should be all-or-nothing, so it shouldn't make a difference (also, `errno` wouldn't be set, so we'd need to make an extra error condition, which seems a bit much for something that should never happen).

net/respondd/src/respondd.c
jplitza: I'm all for milliseconds, then. I really do not want cache files with "5e-10" or something like that to be valid.

NeoRaider:
Thanks for the review, I've updated both pull requests.

NeoRaider:
@jplitza, do you want to have another look at the pull request, or do you think it is fine like this?

jplitza:
Yes, I intended to have another look at both. While this one is probably fine if you considered all my comments, I didn't even get through the whole of the other one. I hope to get it done by the end of the week.

net/respondd/src/respondd.c
```diff
@@ -41,150 +49,395 @@
...
 static void usage() {
-  puts("Usage: respondd [-h] -g <group> -p <port> -i <if0> [-i <if1> ..]");
-  puts("  -g <ip6>         multicast group, e.g. ff02::2:1001");
-  puts("  -p <int>         port number to listen on");
-  puts("  -i <string>      interface on which the group is joined");
-  puts("  -c <string>      Lua command");
-  puts("  -h               this help\n");
+	puts("Usage:");
+	puts("  respondd -h");
+	puts("  respondd [ -p <port> ] [ -g <group> ] -i <if0> [-i <if1> ..] [ -d <dir> ]");
```
jplitza: Sorry, but this still isn't right. Neither is `-i` strictly required (if it is not given, respondd only replies to unicast and broadcast and doesn't join any multicast groups, which I consider a good thing), nor may `-i` be given without `-g` (would output "Multicast group must be given before interface." in line 409). While the other way round (`-g` without `-i`) wouldn't throw an error, it doesn't really make any sense. So I still propose the following:
```
Usage:
  respondd -h
  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir>]
```

net/respondd/README.md
```diff
@@ -0,0 +1,35 @@
+respondd providers are C modules (shared objects). These modules should include
+<respondd.h>, which provides the following definitions:
```
jplitza: "<json-c/json.h> and <respondd.h>, the latter of which provides the following definitions:"

net/respondd/README.md
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        const struct respondd_provider_info respondd_providers[] = {
+                {"nodeinfo", respondd_provider_nodeinfo},
+                {"statistics", respondd_provider_statistics},
+                {}
+        };
+
+The array must be terminated with an empty entry. The provider for each
+request type must return a JSON object. This JSON object must have exactly 1
```
jplitza: I think we definitely need a reference to JSON-C here, as we can not assume everybody is familiar with it:
```
…must return a [JSON-C] JSON object…

[JSON-C]: https://github.com/json-c/json-c/wiki
```

net/respondd/src/respondd.c
NeoRaider: Makes sense.

PR #126, opened 2015-12-27
libplatforminfo

NeoRaider:
Makes the functions of lua-platform-info available from C as well.

The rewritten lua-platform-info is 100% compatible with the old version.


tcatm:
If this really needs to get into the next release (probably in tandem with #125), could we rename the release to 2016.1 and schedule it around March?

NeoRaider:
Well, we can't release 2015.2 in its current state (as the updates are unreliable on 32MB devices...)

I guess we can delay the release for a few weeks, but I'd rather get this out in January than March. Wasn't our previous schedule: Get things done, test for two weeks, release? Just that this time the two weeks' testing showed that it isn't good in its current state...

tcatm:
I've reviewed the code (though, not too carefully) and merging this may be fine. Does this get 2015.2 in a good shape for release or is anything else needed (do we have issues for that?)?

libs/libplatforminfo/src/targets/template/x86.c
```diff
@@ -0,0 +1,107 @@
...
+
+__attribute__((constructor)) static void init(void) {
+        FILE *f = fopen("/proc/cpuinfo", "r");
+        if (!f)
+                return;
+
+        char *line = NULL;
+        size_t len = 0;
+
+        while (getline(&line, &len, f) > 0 && !model) {
```
tcatm: Isn't this just something like `sscanf(line, "model name:%s", &model)` followed by looping using `strtok` to find spaces and tabs and removing those?

libs/libplatforminfo/src/common.h
```diff
@@ -0,0 +1,88 @@
...
+
+__attribute__((unused)) static char * read_line(const char *filename) {
+        FILE *f = fopen(filename, "r");
+        if (!f)
+                return NULL;
+
+        char *line = NULL;
+        size_t len = 0;
+
+        getline(&line, &len, f);
```
tcatm: Why is `len` not used after the call? The `strlen` can probably be avoided.

libs/libplatforminfo/src/common.h
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        return line;
+}
+
+__attribute__((unused)) static void sanitize_image_name(char **outp, char *in) {
+        char *out = malloc(strlen(in) + 1);
+        *outp = out;
+
+        bool dot = false, dash = false;
```
tcatm: Could this use a enum? I.e. `enum {OTHER, DOT, DASH} sep = OTHER;`

NeoRaider: No, this wouldn't handle combinations of dots and dashes correctly.

libs/libplatforminfo/src/common.h
NeoRaider: len is the size of the allocated buffer and doesn't reflect the length of the string.

libs/libplatforminfo/src/targets/template/x86.c
NeoRaider: Well, as long as the colon is always preceded by whitespace, it would work with sscanf, but I didn't want to rely on that (as there are entries in /proc/cpuinfo which don't have whitespace)

NeoRaider:
I'm currently rewriting all respondd providers as C modules, I think I can get that finished today (but I might want to put another day into cleanup).

The new respondd has significatly lower memory consumption, but of course, more tests are necessary to know if it solves our problems. The autoupdater should be rewritten in C as well, but I'd like to delay that till after 2016.1 unless it is necessary to make autoupdates reliable again.

libs/libplatforminfo/src/targets/template/x86.c
NeoRaider: I'll change this to use sscanf, I've just read the manpage again...

NeoRaider: Turns out matching for `%[^\n]` is broken in uClibc, I'll keep the open-coded version for now...

NeoRaider:
I've updated the code to generate a shared library instead of a static one, as having a static library version for each subtarget would cause other issues...

PR #128, opened 2016-01-10
Autoupdater improvements

NeoRaider:
The main change is the addition of the download.d directory, which allows to stop services after the manifest has been verified, but before the image is downloaded.

Suggested list of services to stop (RFC):
* uhttpd
* alfred
* respondd
* batman-adv-visdata
* haveged

### DIFF
--- a/admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
+++ b/admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
@@ -6,8 +6,23 @@ local nixio = require 'nixio'
 module 'autoupdater.util'
 
 
+-- Executes a command in the background, without parsing the command through a shell (in contrast to os.execute)
+function exec(...)
+  local pid, errno, error = nixio.fork()
+  if pid == 0 then
+    nixio.execp(...)
+    os.exit(127)
+  elseif pid > 0 then
+    local wpid, status, code = nixio.waitpid(pid)
+    return wpid and status == 'exited' and code
+  else
+    return pid, errno, error
+  end
+end
+
+
 -- Executes a command in the background, returning its PID and a pipe connected to the command's standard input
-function popen(command)
+function popen(...)
   local inr, inw = nixio.pipe()
   local pid = nixio.fork()
 
@@ -21,7 +36,8 @@ function popen(command)
     inr:close()
     inw:close()
 
-    nixio.exec('/bin/sh', '-c', command)
+    nixio.execp(...)
+    os.exit(127)
   end
 end
 
--- a/admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
+++ b/admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
@@ -1,9 +1,9 @@
-local io = io
-local math = math
 local nixio = require 'nixio'
+local fs = require 'nixio.fs'
+local util = require 'nixio.util'
 
 
-module 'autoupdater.util'
+module('autoupdater.util', package.seeall)
 
 
 -- Executes a command in the background, without parsing the command through a shell (in contrast to os.execute)
@@ -38,6 +38,39 @@ function popen(...)
 
     nixio.execp(...)
     os.exit(127)
+  end
+end
+
+
+-- Executes all executable files in a directory
+function run_dir(dir)
+  local function is_ok(entry)
+    if entry:sub(1, 1) == '.' then
+      return false
+    end
+
+    local file = dir .. '/' .. entry
+    if fs.stat(file, 'type') ~= 'reg' then
+      return false
+    end
+    if not fs.access(file, 'x') then
+      return false
+    end
+
+    return true
+  end
+
+  local files = util.consume(fs.dir(dir))
+  if not files then
+    return
+  end
+
+  table.sort(files)
+
+  for _, entry in ipairs(files) do
+    if is_ok(entry) then
+      exec(dir .. '/' .. entry)
+    end
   end
 end
 
--- a/admin/autoupdater/files/usr/sbin/autoupdater
+++ b/admin/autoupdater/files/usr/sbin/autoupdater
@@ -241,6 +241,10 @@ end
 
 -- Tries to perform an update from a given mirror
 local function autoupdate(mirror)
+  local download_d_dir = '/usr/lib/autoupdater/download.d'
+  local abort_d_dir = '/usr/lib/autoupdater/abort.d'
+
+
   local manifest = read_manifest(mirror)
   if not manifest then
     return false
@@ -259,13 +263,12 @@ local function autoupdate(mirror)
     return true
   end
 
-
-  autoupdater_util.exec('sync')
-  autoupdater_util.exec('sysctl', '-w', 'vm.drop_caches=3')
+  autoupdater_util.run_dir(download_d_dir)
   collectgarbage()
 
   local image = os.tmpname()
   if not fetch_firmware(mirror, manifest.filename, image) then
+    autoupdater_util.run_dir(abort_d_dir)
     return false
   end
 
@@ -275,6 +278,7 @@ local function autoupdate(mirror)
   if checksum ~= manifest.checksum then
     io.stderr:write('Invalid image checksum!\n')
     os.remove(image)
+    autoupdater_util.run_dir(abort_d_dir)
     return false
   end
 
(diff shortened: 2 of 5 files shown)


admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
```diff
@@ -1,13 +1,28 @@
...
 local nixio = require 'nixio'
+local fs = require 'nixio.fs'
+local util = require 'nixio.util'
 
 
-module 'autoupdater.util'
+module('autoupdater.util', package.seeall)
+
+
+-- Executes a command in the background, without parsing the command though a shell (in contrast to os.execute)
```
jplitza: Typo: s/though/through/

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
```diff
@@ -1,13 +1,28 @@
...
+function exec(...)
+  local pid, errno, error = nixio.fork()
+  if pid == 0 then
+    nixio.execp(...)
```
jplitza: According to the [doc](http://luci.subsignal.org/api/nixio/modules/nixio.html#nixio.execp), `execp()` also launches a shell. Why not just `exec()`?

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
```diff
@@ -21,7 +36,41 @@ function popen(command)
     inr:close()
     inw:close()
 
-    nixio.exec('/bin/sh', '-c', command)
+    nixio.execp(...)
```
jplitza: See above.

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
```diff
@@ -21,7 +36,41 @@ function popen(command)
...
+
+-- Executes all executable files in a directory
+function run_dir(dir)
+  local function is_ok(entry)
+    if entry:sub(1, 1) == '.' then
+      return false
+    end
+
+    local file = dir .. '/' .. entry
+    if fs.stat(file, 'type') ~= 'reg' then
```
jplitza: This forbids symlinks. Intended?

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
```diff
@@ -21,7 +36,41 @@ function popen(command)
...
+      return false
+    end
+    if not fs.access(file, 'x') then
+      return false
+    end
+
+    return true
+  end
+
+  local files = util.consume(fs.dir(dir))
```
jplitza: Shouldn't this come after the following if-block?

admin/autoupdater/files/usr/sbin/autoupdater
```diff
@@ -257,21 +259,25 @@ local function autoupdate(mirror)
     return true
   end
 
+  autoupdater_util.run_dir('/usr/lib/autoupdater/download.d')
 
-  os.execute('sync; sysctl -w vm.drop_caches=3')
+  autoupdater_util.exec('sync')
+  autoupdater_util.exec('sysctl', '-w', 'vm.drop_caches=3')
```
jplitza: Why aren't these two commands a script in `download.d`?

admin/autoupdater/files/usr/sbin/autoupdater
```diff
@@ -257,21 +259,25 @@ local function autoupdate(mirror)
...
   collectgarbage()
 
   local image = os.tmpname()
   if not fetch_firmware(mirror, manifest.filename, image) then
+    autoupdater_util.run_dir('/usr/lib/autoupdater/abort.d')
     return false
   end
 
-  local popen = io.popen(string.format("sha512sum '%s'", image))
+  local popen = io.popen(string.format("exec sha512sum '%s'", image))
```
jplitza: Unrelated to the change at hand, but shouldn't we escape `'` in `image`? Or "simply" reimplement `io.popen()` the way we already have our `autoupdater_utils.popen()`, as passing the command as separate arguments would also solve the problem.

In general, I dislike that now we have some commands that are called as separate arguments and some that are still called via a shell as a single string.

admin/autoupdater/files/usr/sbin/autoupdater
```diff
@@ -257,21 +259,25 @@ local function autoupdate(mirror)
...
   local checksum = popen:read('*l'):match('^%x+')
   popen:close()
   if checksum ~= manifest.checksum then
     io.stderr:write('Invalid image checksum!\n')
     os.remove(image)
+    autoupdater_util.run_dir('/usr/lib/autoupdater/abort.d')
```
jplitza: We should probably call `abort.d` around line 300 as well?

jplitza:
Done with the current version of code, waiting for fixes or responses.

Regarding the list of services to stop, this is what `sysupgrade` itself stops anyway on my node:
```
Sending TERM to remaining processes ... netifd crond gluon-crond uradvd uhttpd dnsmasq fastd fastd ntpd alfred batadv-vis dnsmasq respondd ubusd askfirst dhcpv6.script dhcpv6.script dhcpv6.script logd rpcd haveged 
```
Assuming we need to keep everything network-related (netifd, dnsmasq, fastd, dnsmasq, ubusd?, dhcp*), we could stop (in addition to your list):
* crond
* gluon-crond
* uradvd
* ntpd
* ubusd?
* askfirst
* logd?
* rpcd

jplitza:
Also, could you please simply commit the whitespace-fix? It only blows up this PR unnecessarily.

NeoRaider:
I've pushed the whitespace fix.

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
NeoRaider: The docs are plain wrong. The difference between exec and execp is that exec expects an absolute path, while execp respects PATH (just like the execv and execvp C functions internally used by this).

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
NeoRaider: fs.stat gives the information about the file linked to, to get the information about the link itself, fs.lstat must be used.

admin/autoupdater/files/usr/lib/lua/autoupdater/util.lua
NeoRaider: Ah, the following was supposed to be `if not files`, will fix.

admin/autoupdater/files/usr/sbin/autoupdater
NeoRaider: Good idea.

admin/autoupdater/files/usr/sbin/autoupdater
NeoRaider: Indeed. I think making the path configurable is overkill, but having them defined at the top makes sense.

admin/autoupdater/files/usr/sbin/autoupdater
NeoRaider: True, reimplementing io.popen would be the best solution, but seemed like unnecessary work (as the Lua autoupdater will be replaced by C anyways after 2016.1).

The autoupdater_util.exec function, on the other hand, was important, as it is now also used to call the download.d and abort.d executables, which are supplied by other packages and thus might not follow sensible naming conventions ;)

NeoRaider:
Updated branch pushed.

I'd like download.d not to unnecessarily cripple the node. As the download can take very long when the connection to the server is bad, I've kept my list to the monitoring services, so the node will disappear from the map, but apart from that continue to function.

* crond and gluon-crond might be stopped as well, I'm not sure about these. Would have the additional benefit to avoid starting multiple autoupdater instances
* No strong opinion about uradvd or ntpd...
* ubusd and logd are very central and can't be properly brought up in abort.d again once stopped
* askfirst is started by inittab, not an initscript, and is respawned as soon as it is terminated (which is why sysupgrade sends it a SIGKILL...)
* rpcd should not even be installed by default in the Gluon master, we've reverted the LuCI change adding a dependency on rpcd.

jplitza:
Sorry, my master version is a tad older, now that you mention it I remember that change removing rpcd again.

uradvd and ntpd don't really cripple the node: As it seems to have internet/mesh access (it successfully downloaded the manifest) chances are high another router with radvd running is available too. And ntpd isn't even visible to the outside world.

I understand your reservations about ubusd and logd, I had them too and thus marked them with a question mark.

askfirst probably doesn't really matter, but should be safe to somehow kill if we can stop init from restarting it.

PR #130, opened 2016-02-19
autoupdater: add upgrade.d directory.

lemoer:


tcatm:
I think this should be preupgrade.

lemoer:
I would appreciate if you can decide the name of the directory very fast, because I wan't to use the package during an update tomorrow.

NeoRaider:
We've discussed the name on IRC, and decided on "upgrade.d" (as we can't possible have something like "post-upgrade" anyways).

NeoRaider:
(also, download.d is also executed *before* the download of the image)

PR #131, opened 2016-03-19
force autoupdate to ignore release by setting /lib/gluon/release to 0

viisauksena:
so autoupdater will try to do a update - even when it fails in the first place (so next time)
this is nice if wrong release strings are there (for mistakenly have v2106.1.2 for example), or if you do minor changes without having new release string.
(you have to do the autoupdater in this case per hand , but you can profit from the structures .. so you dont have to cre about model and all the stuff)

sideeffect : if wanted you can call autoupdater everytime like this (-F) and it will do everytime a update !!

### DIFF
--- a/admin/autoupdater/files/usr/sbin/autoupdater
+++ b/admin/autoupdater/files/usr/sbin/autoupdater
@@ -30,6 +30,8 @@ version_file:close()
 -- If force is true the updater will perform an upgrade regardless of
 -- the priority and even when it is disabled in uci
 local force = false
+-- if you want to skip release test also
+local superforce = false
 
 -- If fallback is true the updater will perform an update only if the
 -- timespan given by the priority and another 24h have passed
@@ -40,6 +42,9 @@ local function parse_args()
   local i = 1
   while arg[i] do
     if arg[i] == '-f' then
+      force = true
+    elseif arg[i] == '-F' then
+      superforce = true
       force = true
     elseif arg[i] == '--fallback' then
       fallback = true
@@ -252,8 +257,10 @@ local function autoupdate(mirror)
   end
 
   if not autoupdater_version.newer_than(manifest.version, old_version) then
+    if superforce ~= true then
+      return true
+    end
     io.stderr:write('No new firmware available.\n')
-    return true
   end
 
   io.stderr:write('New version available.\n')


jplitza:
Did you read the comment by @NeoRaider in freifunk-gluon/gluon#694? Emptying the file means that even if the update fails, the node will never again know which version it has. That's bad. Instead, simply don't do the version check inside the autoupdater if the flag is set.

tcatm:
This patch has many problems (executing a sub shell, not using version_file). Could you explain why this is needed in more detail?

viisauksena:
i am not as firm in .lua ... 
@jplitza you mean instead suggested something like this in line 44++ 
```
    elseif arg[i] == '-F' then
      superforce = true
```
and line 258 the **if** ...
```
 if not autoupdater_version.newer_than(manifest.version, old_version) then
   if superforce == true then
     return false
   end
   if superforce != true
     return true
   end
   io.stderr:write('No new firmware available.\n')
 end
```
so **stderr** and all general behaviour stay the exactly the same - except you want superforce . (reasons see above)
is it enough here to return false here? 

with this no subshell is needed, no changing of any file is needed ( @tcatm )

viisauksena:
hope this is right, i testest this by hand and it seems to work as intendet

admin/autoupdater/files/usr/sbin/autoupdater
```diff
@@ -41,6 +43,9 @@ local function parse_args()
   while arg[i] do
     if arg[i] == '-f' then
       force = true
+    elseif arg[i] == '-F' then
```
NeoRaider: As mentioned in the other bug report, I'd prefer a more verbose flag here, something like `--force-version` or `--ignore-version`.

admin/autoupdater/files/usr/sbin/autoupdater
```diff
@@ -252,8 +257,10 @@ local function autoupdate(mirror)
   end
 
   if not autoupdater_version.newer_than(manifest.version, old_version) then
```
NeoRaider: Just change the condition to `if not superforce and not autoupdater_version.newer_than(manifest.version, old_version) then` (and you might want to adjust the messages to be less confusing then). Also, the variable should not be called `superforce`, but match whatever option name we choose.

NeoRaider:
Some general comments:
* I'm not opposed to merging this, but @tcatm asked for an explanation of your usecase
* You need to squash your commits into one
* As mentioned in the other ticket, the Lua autoupdater will be replaced completely before the next Gluon release, so I'm not sure if it makes any sense to add a new feature to it now...

viisauksena:
@tcatm @NeoRaider usecases:
* you maybe do not want to have a extra release in the wild for minor changes - but bring it to specific nodes
* you maybe want to benefit from working autoupdater in terms of finding correctly signed firmware for your model and just do a sysupgrade even if it thing it has the newest version already (this is what i believed force would do - but i am ok, that it does not)
* you may have messed a release with wrong release number - like typo v2106.1.2 , autopupdater -f think it has already the latest.
* you may want to easily test autoupdate procedures
.. maybe all minor - but so easy to "fix" or have.

the squash will be done 

@NeoRaider where do i find info/source of new autoupdater - and, will this be buildin (as a real -f = force in the first place or a -F superforce as is called it)

admin/autoupdater/files/usr/sbin/autoupdater
viisauksena: ah, read it too late .. wont make any big difference, while autoupdater will be replaced - as you said

admin/autoupdater/files/usr/sbin/autoupdater
viisauksena: i could change it, but as you said the autpupdater will be replaced - if you say : do it and i merge it , ok
otherwise, you could merge it already

rotanid:
closing PR as the author didnt do the requested change(s) and there will be a new autoupdater.

viisauksena:
@rotanid  : what have i missed to do? , i sqashed the commits and i explained some of the usecases - one extra would be : if you change branches, the release numbers may not match up

rotanid:
@viisauksena there were two additional request for changes in your code, just look above. feel free to ping me again as soon as you implemented those.

PR #132, opened 2016-04-08
lua-hash: introduce a lua hashing library using libubox.

lemoer:
references freifunk-gluon/gluon#715

### DIFF
--- a/libs/lua-hash/src/hash.c
+++ b/libs/lua-hash/src/hash.c
@@ -1,0 +1,47 @@
+#include <libubox/md5.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define MD5_DIGEST_LEN 16
+#define MD5_STRLEN (2*MD5_DIGEST_LEN+1)
+
+static int md5(lua_State *L) {
+	size_t len;
+	const char *input = lua_tolstring(L, 1, &len);
+	if (input == NULL) {
+		lua_pushstring(L, "first argument: should be the input string");
+		lua_error(L);
+		__builtin_unreachable();
+	}
+
+	unsigned char digest[MD5_DIGEST_LEN];
+	char output[MD5_STRLEN];
+
+	struct md5_ctx ctx;
+	md5_begin(&ctx);
+	md5_hash(input, len, &ctx);
+	md5_end(digest, &ctx);
+
+	// fill the digest bytewise in the output string
+	int i;
+	for (i = 0; i < MD5_DIGEST_LEN; i++) {
+		sprintf(output + 2*i, "%02x", digest[i]);
+	}
+
+	lua_pushstring(L, output);
+	return 1;
+}
+
+static const luaL_Reg R[] = {
+	{"md5", md5},
+	{NULL, NULL },
+};
+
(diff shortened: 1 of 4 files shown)


NeoRaider:
Looks quite good, I only have a few comments.

libs/lua-hash/src/hash.c
```diff
@@ -0,0 +1,47 @@
...
+
+#define MD5_DIGEST_LEN 16
+#define MD5_STRLEN 2*MD5_DIGEST_LEN+1
+
+static int md5(lua_State *L) {
+	char *input = (char *) lua_tostring(L, 1);
+	if (input == NULL) {
+		lua_pushstring(L, "first argument: should be the input string");
+		lua_error(L);
+		return 1;
```
NeoRaider: The return is superfluous, as lua_error will longjmp. It should probably be replaced with `__builtin_unreachable()`, as Lua doesn't define the function with attribute((noreturn)).

libs/lua-hash/src/hash.c
```diff
@@ -0,0 +1,47 @@
...
+#include <lauxlib.h>
+#include <lualib.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define MD5_DIGEST_LEN 16
+#define MD5_STRLEN 2*MD5_DIGEST_LEN+1
+
+static int md5(lua_State *L) {
+	char *input = (char *) lua_tostring(L, 1);
```
NeoRaider: This should use luaL_checklstring (so only strings are accepted); the "lstring" version has an additional argument for the length, so strings with embedded zeros can be supported.

The cast chould be removed, *input* should be a `const char *`.

libs/lua-hash/src/hash.c
```diff
@@ -0,0 +1,47 @@
...
+
+	struct md5_ctx ctx;
+	md5_begin(&ctx);
+	md5_hash(input, strlen(input), &ctx);
+	md5_end(digest, &ctx);
+
+	// fill the digest bytewise in the output string
+	int i;
+	for (i = 0; i < MD5_DIGEST_LEN; i++) {
+		sprintf(output + 2*i, "%02x", (unsigned char) digest[i]);
```
NeoRaider: Instead of casting to unsigned char here, you could just declare *digest* as an unsigned char array.

libs/lua-hash/src/hash.c
```diff
@@ -0,0 +1,47 @@
...
+	}
+	output[MD5_STRLEN - 1] = '\0';
```
NeoRaider: sprintf will always terminate the string, to this line can be removed.

lemoer:
@NeoRaider Is the code now ready for merging? Then I would do a squash commit.

libs/lua-hash/src/hash.c
```diff
@@ -0,0 +1,47 @@
+#include <libubox/md5.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define MD5_DIGEST_LEN 16
+#define MD5_STRLEN 2*MD5_DIGEST_LEN+1
```
tcatm: I'd prefer this inlined (or at least wrapped in parantheses).

NeoRaider: I agree, I overlooked this in my review.

PR #133, opened 2016-04-24
basic package for l3roamd

christf:


christf:
- [ ] add init script
- [ ] add configuration (will l3roamd be able to recognize interface changes?)

christf:
There is no initscript. gluon-l3roamd should handle initalization.

PR #134, opened 2016-04-27
uradvd: make AdvDefaultLifetime configurable

tcatm:
Introduce a new flag --default-lifetime <seconds>.

### DIFF
--- a/net/uradvd/src/uradvd.c
+++ b/net/uradvd/src/uradvd.c
@@ -30,6 +30,8 @@
 #include <error.h>
 #include <ifaddrs.h>
 #include <fcntl.h>
+#include <getopt.h>
+#include <limits.h>
 #include <poll.h>
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


net/uradvd/src/uradvd.c
```diff
@@ -583,8 +588,30 @@ static void add_prefix(const char *prefix, bool adv_onlink) {
...
+
+	int option_index = 0;
+
+	while ((c = getopt_long(argc, argv, "i:a:p:h", long_options, &option_index)) != -1) {
 		switch(c) {
+		case 0: // --default-lifetime
+			errno = 0;
+			val = strtoul(optarg, &endptr, 0);
+
+			if (((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN)) || (errno != 0 && val == 0) || endptr == optarg) && val <= UINT16_MAX)
```
NeoRaider: The `(val == LONG_MAX || val == LONG_MIN)` doesn't make sense for strtoul, and there are other mistakes in the conditoin. AFAICT, the following check would suffice:

    if (!*optarg || *endptr || val > UINT16_MAX)

*errno* can be ignored completely.

PR #135, opened 2016-05-04
add --rdnss option

tcatm:
This option may be specified multiple times (up to 8).


jplitza:
FTR: Actually, you can specify it up to 3 times, not 8.

PR #136, opened 2016-05-06
Add a sysupgrade-watchdog

RubenKelevra:


tcatm:
Thanks for the idea.

This needs:

- a better commit message
- add documentation

Also, can it use the kernel watchdog feature? What will happen if the watchdog triggers (due to some bad luck) while rewriting the flash?

NeoRaider:
Also, how can this even happen? I've sometimes seen the autoupdater getting stuck due to wget (which I'm currently solving be rewriting the autoupdater), but sysupgrade is a process that should be too simple to fail (and I agree with tcatm that any kind of interruption is too dangerous to do automatically)

RubenKelevra:
We had some issues some while ago, on which sysupgrade seems to be interrupted after flashing, so if we reboot it which this watchdog, the nodes does not stuck after flashing the flash.

Since the flash-process needs only some seconds, the timeout is big enough by far, but if you still bother, we could simply double it.

Since this is the last resort for a reboot I'm not sure if a kernel-watchdog would do the trick as well. On our nodes this hack successfully resolves any issue while updating. So we decided to send it upstream.

Adorfer:
As a matter of fact i see a 0,5-1% of nodes which get stuck during autoupdate. 
They freeze until beeing powercycled. And then start with the old FW.
I do not know how they crash, but i will give this a try.

Adorfer:
Also zumindest so verhindert es sicher ein Upgrade auf Futro und 841ern.

Adorfer:
obiges Script aus dem PR führt dazu, dass die Subshell in einem "sleep 400" wartet während des upgrades. Und dann besorgt der Kill-Job des Sysupgrade, dass das Sleep ein SigTerm bekommt... Reboot... bevor der Flasher fertig ist. 
Das ist mindestens kontraproduktiv.

Vermutlich x-fach doppelt und/ineffektiv. Aber funktioniert so:

> 
> #!/bin/sh
> #watchdog which inits a reboot if sysupgrade is stuck
> echo starting 99watchdog
> logger starting 99watchdog
> (trap "echo ignoring SIGHUP" 1 
>  trap "echo ignoring SIGINT" 2
>  trap "echo ignoring SIGTERM" 15 
>  trap "echo ignoring SIGKILL" 9 
>  sleep 500; sleep 400; sleep 300; sleep 200;reboot&>/dev/null;reboot)&

NeoRaider:
Please provide logs if there's something wrong with upgrades. It is too easy to brick devices during upgrades to play around with half-baked workarounds when we don't even know what is happening.

rotanid:
no answer, closing.
feel free to contribute the information that was asked for and we'll reopen this issue.

PR #137, opened 2016-05-06
Create a lock-file while running autoupdate-process

RubenKelevra:
There are many scripts out there, which detect some error-states of the freifunk-router, which actually conflicts with the autoupdate-process. So some might detect issues while updating and would init a reboot while sysupgrade is still writing the image.

To ensure such race conditions does not occur this patch provides a simple lock-file which can be checked by 3rd-party software to ensure autoupdate is not running before they init a reboot.

### DIFF
--- a/admin/autoupdater/files/usr/lib/autoupdater/upgrade.d/00lockfile
+++ b/admin/autoupdater/files/usr/lib/autoupdater/upgrade.d/00lockfile
@@ -1,0 +1,8 @@
+# Create a lockfile for 3rd-party scripts which can be checked easily to ensure that
+# no reboot will be initiated while autoupdater is running
+
+lockfile='/tmp/autoupdate.lock'
+
+touch $lockfile || logger "gluon-autoupdater: lockfile could not be written"
+
+exit 0


Adorfer:
Ich weiss ncht, warum einige Router währenend des Updates nach wie vor hängen bleiben. Es sind gefühlt um die 1% bei jedem Lauf. 
Aber da die sich dann mit einem einfachen Powercycle wieder (auf die vorherige Version) besinnen und der nächste Versuch dann funktioniert, werde ich diesem Patch mal eine Chance geben.

jplitza:
1. Why not simply check in the 3rd-party software whether a process with the name `sysupgrade` is running?
2. Your approach could be suspect to TOCTOU:
  1. 3rd-party software checks if lockfile exists. It doesn't.
  2. autoupdater creates lockfile and starts sysupgrade
  3. 3rd-party software reboots

One possible way around the TOCTOU and really the only reason for a lockfile at all, is not to run the upgrade if the lockfile exists. Then, 3rd-party software could create the lockfile and proceed. Note that creating the lockfile, both by the autoupdater and the 3rd-party software, should then be done by `open(..., O_EXCL|O_CREAT)`.

RubenKelevra:
@Adorfer für den Bug hab ich ja das hier:

https://github.com/freifunk-gluon/packages/pull/136

@jplitza 

Well we have some fixing scripts which checks this lock file immediately before hitting the reboot button. Since it still could cause a race condition when the script checks for "sysupgrade" and then hitting reboot while autoupdater is starting up sysupgrade, this patch here seems far more reliable.

This patch does not hurt anybody but provide a simple way to ensure both scripts does not interference.

jplitza:
@RubenKelevra I don't see why this is in any way less prone to race conditions.

RubenKelevra:
If I create a file before starting up a process, which also takes some seconds to initialize is safe.

Checking for a process running is far more complex on the one hand and depends on a already started process, so this is less safe.

While a lockfile just need a simple "[ -f $lockfile ] && exit 1" checking for a running process with a given name does need some external tools, doesn't it?

NeoRaider:
The autoupdater is already trying to stop everything that might get in the way (and sysuprade should kill the rest). Instead of adding a race-prone lockfile, just make sure that relevant software is stopped by the autoupdater (it should also be easer to just add a `/etc/init.d/... stop` call in one of the `*.d` directories of the autoupdater than it would be to respect some random lock file...)

NeoRaider:
The autoupdater will stop almost everything, *sysupgrade* will send SIGTERM and SIGKILL to the rest, so this is not really an issue. Rebooting while an upgrade runs can't really happen.

PR #139, opened 2016-06-18
Doku zu respondd hinzugefügt

yanosz:
Hoffe, wir haben das soweit verstanden.

### DIFF
--- a/net/respondd/README.md
+++ b/net/respondd/README.md
@@ -1,3 +1,109 @@
+# What is respondd?
+
+## Introduction
+respondd is a server distributing information within a network.
+
+For doing so, respondd spawns a udp socket (1001/udp) joining a multicast group. When a request is received, the information requested is transmitted to the requestor. 
+
+All information is oranized in a non-hierachical namespace. Each entry identifies a *provider* implemented in at least one C module. (eg `statistics`, `nodeinfo`, `...`) - each request queries a specific provider. 
+Each response is the merge-result of the output of all modules registered for the provider.
+
+Some providers are implemented in [/package/gluon-respondd](https://github.com/freifunk-gluon/gluon/tree/master).
+## Usage
+
+TODO: Why is the port number included in the multicast group, too?
+```
+respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir>]
+  -p <int>         port number to listen on
+  -g <ip6>         multicast group, e.g. ff02::2:1001
+  -i <string>      interface on which the group is joined
+  -d <string>      data provider directory (default: current directory)
+  -h               this help
+```
+
+
+
+## Procotol
+
+Request and response are encoded as *utf8*-strings. These strings are sent as UDP-Packets. Fragmentation is not supported -  gzip-compression (for the response) is optional. 
+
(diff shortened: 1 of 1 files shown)


net/respondd/README.md
```diff
@@ -1,3 +1,109 @@
...
+
+For doing so, respondd spawns a udp socket (1001/udp) joining a multicast group. When a request is received, the information requested is transmitted to the requestor. 
+
+All information is oranized in a non-hierachical namespace. Each entry identifies a *provider* implemented in at least one C module. (eg `statistics`, `nodeinfo`, `...`) - each request queries a specific provider. 
+Each response is the merge-result of the output of all modules registered for the provider.
+
+Some providers are implemented in [/package/gluon-respondd](https://github.com/freifunk-gluon/gluon/tree/master).
+## Usage
+
+TODO: Why is the port number included in the multicast group, too?
```
NeoRaider: The :1001 is not the port, it's just an arbitrary part of the multicast group address.

NeoRaider:
Looks very good. It might be added that the whole multicast part is optional, and respondd also also answers unicast requests.

net/respondd/README.md
```diff
@@ -1,3 +1,109 @@
...
+  -i <string>      interface on which the group is joined
+  -d <string>      data provider directory (default: current directory)
+  -h               this help
+```
+
+
+
+## Procotol
+
+Request and response are encoded as *utf8*-strings. These strings are sent as UDP-Packets. Fragmentation is not supported -  gzip-compression (for the response) is optional.
```
NeoRaider: This is slightly off - respondd uses *deflate* compression, without gzip header.

yanosz:
Feel free to modify & merge. Guess you know the tool best ;-)

net/respondd/README.md
```diff
@@ -1,3 +1,109 @@
...
+  -h               this help
+```
+
+
+
+## Procotol
+
+Request and response are encoded as *utf8*-strings. These strings are sent as UDP-Packets. Fragmentation is not supported -  gzip-compression (for the response) is optional. 
+
+- The request is the name for the provider - '`GET `' can be prepended for compression
```
jplitza: `GET` not only enables compression, but also changes the output format and allows for multiple providers to be requested. So for example you can use `GET nodeinfo statistics` to get both providers in response to one request.

net/respondd/README.md
```diff
@@ -1,3 +1,109 @@
...
+Requesting `nodeinfo` as implemented in the freifunk-gluon modules.
+
+#### Request
+```
+nodeinfo
+```
+
+#### Response
+```json
+"nodeinfo": {
```
jplitza: This isn't actually valid JSON (and thus probably not copied from an actual output?). You are missing a surrounding pair of `{}`. And as mentioned above, this is actually the (decompressed) output of a `GET nodeinfo` request, a pure request of `nodeinfo` wouldn't have wrapped the return data in another object with the key `nodeinfo`.

NeoRaider:
I'll pull it soon with a few modifications.

NeoRaider:
I've pulled your update and pushed it with some changes.

PR #143, opened 2016-07-30
respondd: Read list of multicast interfaces from file, add initscript

jplitza:
We currently restart respondd regularly if something about the interfaces changes. This makes it possible to instead send a SIGHUP and respondd will reload the list of interfaces from a file.

Furthermore, if we ever want to upstream respondd (or make it usable outside Gluon), an initscript is more than helpful. Thus, I added one that uses UCI variables for all parameters and procd to log the stderr.

### DIFF
--- a/net/respondd/Makefile
+++ b/net/respondd/Makefile
@@ -25,6 +25,7 @@ define Build/Prepare
 endef
 
 define Package/respondd/install
+	$(CP) ./files/* $(1)/
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/respondd $(1)/usr/bin/
 endef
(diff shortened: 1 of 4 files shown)


rotanid:
this would be nice, we have some nodes where respondd is crashing and this init script might start it again after the crash

NeoRaider:
I think I'd like to merge #144 first, so this will need to be rebased.

rotanid:
update: as #144 is now merged, this isn't blocked anymore but has to be rebased.

jplitza:
Yes, and heavily so. Because now, not only the interface list needs to be passed, but also the delay information. I don't yet have an idea how to handle that.

jplitza:
This is basically obsoleted by freifunk-gluon/gluon#972, so I'm closing this.

If we wanted to avoid restarts, @NeoRaider indicated that the preferred solution would be to simply give all possible interfaces on the first invocation and listen for netlink events regarding interface state changes inside the daemon.

PR #144, opened 2016-08-13
respondd: Implement delay to multicast requests

lemoer:
I tried to implement a delay to multicast requests as mentioned in #140.

What do think about it?

### DIFF
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -501,20 +501,7 @@ int main(int argc, char **argv) {
 	struct sockaddr_in6 server_addr = {};
 	struct in6_addr mgroup_addr;
 
-	// get radom seed
-	int frandom = open("/dev/urandom", O_RDONLY);
-	if (frandom < 0) {
-		perror("opening random");
-		exit(EXIT_FAILURE);
-	}
-
-	unsigned int seed;
-	if(read(frandom, &seed, sizeof(seed)) < 0) {
-		perror("reading random");
-		exit(EXIT_FAILURE);
-	}
-	close(frandom);
-	srand(seed);
+	srand(time(NULL));
 
 	sock = socket(PF_INET6, SOCK_DGRAM, 0);
 
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -196,7 +196,7 @@ bool queue_push_request(struct request_queue *q, char* req,
 	q->push_task->client_addrlen = addrlen;
 	q->push_task->unprocessed = true;
 
-	if (q->push_task++ > &q->task_ring[QUEUE_RING_LEN-1])
+	if (++q->push_task > &q->task_ring[QUEUE_RING_LEN-1])
 		q->push_task = &q->task_ring[0];
 
 	return true;
@@ -212,7 +212,7 @@ struct request_task* queue_pop_request(struct request_queue *q) {
 	current_task->unprocessed = false;
 
 	// go on to next task
-	if (q->pop_task++ > &q->task_ring[QUEUE_RING_LEN-1])
+	if (++q->pop_task > &q->task_ring[QUEUE_RING_LEN-1])
 		q->pop_task = &q->task_ring[0];
 
 	return current_task;
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -216,13 +216,27 @@ bool schedule_push_request(struct request_schedule *q, char* req,
 	return true;
 }
 
+int64_t schedule_idle_time(struct request_schedule *schedule) {
+	if (!schedule->list_head)
+		// nothing to do yet, wait nearly infinite time
+		return 3600000;
+
+	update_time();
+	int64_t result = schedule->list_head->scheduled_time - now;
+
+	if (result < 0)
+		return 0;
+	else
+		return result;
+}
+
 // the returned task is already set as processed
 struct request_task * schedule_pop_request(struct request_schedule *q) {
 	if (!q->list_head)
 		// schedule is empty
 		return NULL;
 
-	if (q->list_head->scheduled_time > now) {
+	if (schedule_idle_time(q) > 0) {
 		// nothing to do yet
 		return NULL;
 	}
@@ -388,11 +402,19 @@ static struct json_object * handle_request(char *request, bool *compress) {
 }
 
 // the return value indicates whether there was a request
-static bool accept_request(struct request_schedule *schedule, int sock) {
+static bool accept_request(struct request_schedule *schedule, int sock, uint64_t timeout) {
 	char input[REQUEST_MAXLEN];
 	ssize_t input_bytes;
 	struct sockaddr_in6 addr;
 	socklen_t addrlen = sizeof(addr);
+
+	// set timeout to the socket
+	struct timeval t;
+	t.tv_sec = timeout / 1000;
+	t.tv_usec = (timeout % 1000) * 1000;
+
+	if (setsockopt (sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&t, sizeof(t)) < 0)
+		perror("setsockopt failed\n");
 
 	input_bytes = recvfrom(sock, input, sizeof(input)-1, 0, (struct sockaddr *)&addr, &addrlen);
 
@@ -408,6 +430,7 @@ static bool accept_request(struct request_schedule *schedule, int sock) {
 	input[input_bytes] = 0;
 
 	// TODO: only multicast requests should be scheduled in future
+	update_time();
 	schedule_push_request(schedule, input, (struct sockaddr *)&addr, addrlen, now + 5000);
 	return true;
 }
@@ -485,14 +508,6 @@ int main(int argc, char **argv) {
 		exit(EXIT_FAILURE);
 	}
 
-	struct timeval timeout;
-	timeout.tv_sec = 0;
-	timeout.tv_usec = 10000;
-
-	if (setsockopt (sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&timeout,
-	    sizeof(timeout)) < 0)
-		perror("setsockopt failed\n");
-
 	server_addr.sin6_family = AF_INET6;
 	server_addr.sin6_addr = in6addr_any;
 
@@ -551,9 +566,7 @@ int main(int argc, char **argv) {
 	struct request_schedule schedule = {};
 
 	while (true) {
-		update_time();
-		// TODO: adjust timeout, remove polling
-		accept_request(&schedule, sock);
+		accept_request(&schedule, sock, schedule_idle_time(&schedule));
 		serve_request(&schedule, sock);
 	}
 
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -391,6 +391,7 @@ static struct json_object * handle_request(char *request, bool *compress) {
 //     until we reach the scheduling deadline.
 // 2a. If the incomming request was sent to a multicast destination IPv6,
 //     choose a random delay between 0 and max_multicast_delay milliseconds.
+// 2b. If the schedule is full, send the reply immediately.
 // 2c. If the incomming request was sent to a unicast destination, the response
 //     will be also sent immediately.
 static void accept_request(struct request_schedule *schedule, int sock,
@@ -468,6 +469,8 @@ static void accept_request(struct request_schedule *schedule, int sock,
 	}
 
 	if (!schedule_push_request(schedule, new_task)) {
+		// we can't schedule, so send the response immediately
+		serve_request(new_task, sock);
 		free(new_task);
 	}
 }
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -617,7 +617,7 @@ int main(int argc, char **argv) {
 	while (true) {
 		accept_request(&schedule, sock, max_multicast_delay);
 
-		struct request_task* task = schedule_pop_request(&schedule);
+		struct request_task *task = schedule_pop_request(&schedule);
 
 		if (!task)
 			continue;
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -391,8 +391,8 @@ static struct json_object * handle_request(char *request, bool *compress) {
 //     until we reach the scheduling deadline.
 // 2a. If the incomming request was sent to a multicast destination IPv6,
 //     choose a random delay between 0 and max_multicast_delay milliseconds.
-// 2b. If the incomming request was sent to a unicast destination, delay
-//     is zero.
+// 2c. If the incomming request was sent to a unicast destination, the response
+//     will be also sent immediately.
 static void accept_request(struct request_schedule *schedule, int sock,
                            uint64_t max_multicast_delay) {
 	char input[REQUEST_MAXLEN];
@@ -454,19 +454,18 @@ static void accept_request(struct request_schedule *schedule, int sock,
 
 	input[input_bytes] = 0;
 
-	// only multicast requests are delayed
-	int64_t delay;
-	if (IN6_IS_ADDR_MULTICAST(&destaddr)) {
-		delay = rand() % max_multicast_delay;
-	} else {
-		delay = 0;
-	}
-
 	struct request_task *new_task = malloc(sizeof(*new_task));
-	new_task->scheduled_time = now + delay;
+	new_task->scheduled_time = now + rand() % max_multicast_delay;
 	strncpy(new_task->request, input, input_bytes + 1);
 	new_task->request[input_bytes] = 0;
 	new_task->client_addr = addr;
+
+	if (!IN6_IS_ADDR_MULTICAST(&destaddr)) {
+		// unicast packets are sent directly (delay is ignored)
+		serve_request(new_task, sock);
+		free(new_task);
+		return;
+	}
 
 	if (!schedule_push_request(schedule, new_task)) {
 		free(new_task);
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -398,7 +398,6 @@ static void accept_request(struct request_schedule *schedule, int sock,
 	char input[REQUEST_MAXLEN];
 	ssize_t input_bytes;
 	struct sockaddr_in6 addr;
-	socklen_t addrlen = sizeof(addr);
 	char control[256];
 	struct in6_addr destaddr;
 	struct cmsghdr *cmsg;
@@ -467,7 +466,7 @@ static void accept_request(struct request_schedule *schedule, int sock,
 	new_task->scheduled_time = now + delay;
 	strncpy(new_task->request, input, input_bytes + 1);
 	new_task->request[input_bytes] = 0;
-	memcpy(&new_task->client_addr, &addr, addrlen);
+	new_task->client_addr = addr;
 
 	if (!schedule_push_request(schedule, new_task)) {
 		free(new_task);
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -95,7 +95,7 @@ static void usage() {
 	puts("        -p <int>         port number to listen on");
 	puts("        -g <ip6>         multicast group, e.g. ff02::2:1001");
 	puts("        -i <string>      interface on which the group is joined");
-	puts("        -t <int>         delay seconds before multicast responses (default: 10)");
+	puts("        -t <int>         maximum delay seconds before multicast responses (default: 10)");
 	puts("        -d <string>      data provider directory (default: current directory)");
 	puts("        -h               this help\n");
 }
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -187,19 +187,15 @@ bool schedule_push_request(struct request_schedule *s, struct request_task *new_
 		// schedule is full
 		return false;
 
-	if (!s->list_head || s->list_head->scheduled_time > new_task->scheduled_time) {
-		new_task->next = s->list_head;
-		s->list_head = new_task;
-	} else {
-		struct request_task *t;
-		for (t = s->list_head; t && t->next; t = t->next) {
-			if (t->next->scheduled_time > new_task->scheduled_time) {
-				break;
-			}
-		}
-		new_task->next = t->next;
-		t->next = new_task;
+	// insert into sorted list
+	struct request_task **pos;
+	for (pos = &s->list_head; *pos; pos = &((*pos)->next)) {
+		if ((*pos)->scheduled_time > new_task->scheduled_time)
+			break;
 	}
+	// insert before *pos
+	new_task->next = *pos;
+	*pos = new_task;
 
 	s->length++;
 	return true;
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -208,7 +208,7 @@ int64_t schedule_idle_time(struct request_schedule *s) {
 	update_time();
 	int64_t result = s->list_head->scheduled_time - now;
 
-	if (result == 0)
+	if (result <= 0)
 		return -1; // zero is infinity
 	else
 		return result;
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -49,7 +49,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 
-#define QUEUE_RING_LEN 8
+#define SCHEDULE_LEN 8
 #define REQUEST_MAXLEN 256
 
 struct provider_list {
@@ -77,7 +77,7 @@ struct request_task {
 	char request[REQUEST_MAXLEN];
 };
 
-struct request_queue {
+struct request_schedule {
 	int length;
 	struct request_task *list_head;
 };
@@ -184,12 +184,11 @@ static const struct respondd_provider_info * get_providers(const char *filename)
 	return ret;
 }
 
-bool queue_push_request(struct request_queue *q, char* req,
+bool schedule_push_request(struct request_schedule *q, char* req,
                         struct sockaddr *addr, socklen_t addrlen,
                         int64_t scheduled_time) {
-	// TODO: rename QUEUE_RING_LEN
-	if (q->length >= QUEUE_RING_LEN)
-		// queue is full
+	if (q->length >= SCHEDULE_LEN)
+		// schedule is full
 		return false;
 
 	struct request_task *new_task = malloc(sizeof(struct request_task));
@@ -217,9 +216,9 @@ bool queue_push_request(struct request_queue *q, char* req,
 }
 
 // the returned task is already set as processed
-struct request_task * queue_pop_request(struct request_queue *q) {
+struct request_task * schedule_pop_request(struct request_schedule *q) {
 	if (!q->list_head)
-		// queue is empty
+		// schedule is empty
 		return NULL;
 
 	if (q->list_head->scheduled_time > now) {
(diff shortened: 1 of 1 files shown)


lemoer:
Pushed some additional improvements and fixes. The last commits were late in the night... ;)

lemoer:
It seems that the decreased loss is definitively visible:

Grafana Statistic

(The patch was applied at 08/16 around 2.00h)

The collecting tool was [node-informant](https://github.com/ffdo/node-informant).

rotanid:
i cant comment on the code, but before it can be merged it has to be squashed into one single commit instead of 21 ;)

lemoer:
@rotanid Yes. I think this can be done when the feature will be [merged](https://help.github.com/articles/about-pull-request-merge-squashing/). I pushed it that way to allow others to follow how this feature was developed. I'm not sure, whether it is a good practice to do so?

jplitza:
Let's discuss the need of squashing in freifunk-gluon/gluon#865

net/respondd/src/respondd.c
```diff
@@ -320,27 +390,92 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
 	if (input_bytes < 0) {
-		perror("recvfrom failed");
+		perror("recvmsg failed");
 		exit(EXIT_FAILURE);
 	}
 
+	// determine destination address
+	for (cmsg = CMSG_FIRSTHDR(&mh);	cmsg != NULL; cmsg = CMSG_NXTHDR(&mh, cmsg))
+	{
+		// ignore the control headers that don't match what we WARRANTIES
```
jplitza: > what we WARRANTIES

? o_O

net/respondd/src/respondd.c
```diff
@@ -419,6 +584,14 @@ int main(int argc, char **argv) {
 			join_mcast(sock, mgroup_addr, optarg);
 			break;
 
+		case 't':
+			max_multicast_delay = 1000 * atoi(optarg);
```
jplitza: Better use `strtoul()`. That gives you the positivity check for free and lets you catch strings that are not numbers and thus invalid. (Yes, the same would apply for `-p` above)

net/respondd/src/respondd.c
```diff
@@ -57,13 +60,27 @@ struct provider_list {
...
 	struct json_object *cache;
 	uint64_t cache_time;
 	int64_t cache_timeout;
 };
 
+struct request_task {
+	struct request_task *next;
+	int64_t scheduled_time;
+
+	struct sockaddr_storage client_addr;
```
jplitza: Why not use `sockaddr_in6` here like everywhere else in the code?

net/respondd/src/respondd.c
```diff
@@ -165,6 +183,60 @@ static const struct respondd_provider_info * get_providers(const char *filename)
 	return ret;
 }
 
+bool schedule_push_request(struct request_schedule *s, struct request_task *new_task) {
+	if (s->length >= SCHEDULE_LEN)
+		// schedule is full
+		return false;
+
+	if (!s->list_head || s->list_head->scheduled_time > new_task->scheduled_time) {
```
jplitza: By storing `t->next` in a `struct request_task **`, you could get rid of the if/else. Probably a matter of taste, but I'd like to avoid the additional level of indentation (see [here](https://github.com/freifunk-gluon/packages/pull/143/files#diff-ba51ba086696fdaa98da33f36b0beb50R154))

net/respondd/src/respondd.c
```diff
@@ -320,27 +390,92 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+		.iov_base = input,
+		.iov_len = sizeof(input) - 1
+	};
+
+	struct msghdr mh = {
+		.msg_name = &addr,
+		.msg_namelen = sizeof(addr),
+		.msg_iov = &iv,
+		.msg_iovlen = 1,
+		.msg_control = control,
```
jplitza: I suppose `.msg_control = NULL` and `.msg_controllen = 0` would work too, as you are never using the control data again.

net/respondd/src/respondd.c
```diff
@@ -165,6 +183,60 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+	}
+
+	s->length++;
+	return true;
+}
+
+int64_t schedule_idle_time(struct request_schedule *s) {
+	if (!s->list_head)
+		// nothing to do yet, wait nearly infinite time
+		return 3600000;
```
jplitza: This should be `return 0;`, which according to `socket(7)` means "never timeout".

net/respondd/src/respondd.c
```diff
@@ -165,6 +183,60 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+
+	update_time();
+	int64_t result = s->list_head->scheduled_time - now;
+
+	if (result < 0)
+		return 0;
```
jplitza: This, in contrast, probably should be 1 (which blocks for only one millisecond) or something negative (which would have to be handled elsewhere).

net/respondd/src/respondd.c
```diff
@@ -320,27 +390,92 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
 }
 
-
-static void serve(int sock) {
-	char input[256];
-	const char *output = NULL;
-	ssize_t input_bytes, output_bytes;
+// the return value indicates whether there was a request
+static bool accept_request(struct request_schedule *schedule, int sock,
+                           uint64_t timeout, uint64_t max_multicast_delay) {
```
jplitza: I think it makes more sense to calculate the timeout inside this function, as it is computable from the schedule which this function has to have access to anyway.

net/respondd/src/respondd.c
```diff
@@ -165,6 +183,60 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+		}
+		new_task->next = t->next;
+		t->next = new_task;
+	}
+
+	s->length++;
+	return true;
+}
+
+int64_t schedule_idle_time(struct request_schedule *s) {
```
jplitza: The return type should be unsigned (except if you decide to return something negative on error). Currently, it is silently cast to unsigned in `main()` and never receives any negative values anyway.

net/respondd/src/respondd.c
```diff
@@ -388,15 +547,21 @@ int main(int argc, char **argv) {
...
+		exit(EXIT_FAILURE);
+	}
+
 	server_addr.sin6_family = AF_INET6;
 	server_addr.sin6_addr = in6addr_any;
 
 	opterr = 0;
 
 	int group_set = 0;
+	int max_multicast_delay = 10000;
```
jplitza: I think 10 seconds default are a bit high. Do lower values, like maybe 5 seconds, yield similar results?

jplitza:
Haven't compiled yet, but except for some minor things (see my line comments) the code looks good to me.

net/respondd/src/respondd.c
lemoer: I'm not sure about the best value for the maximum delay. The current test does not say anything about the behavior in the real scenario.

Some more facts about the graph:
* I have deployed the patch to a few routers (experimental branch) in the freifunk hanover community and the remaining routers are still answering to the multicast request immediately. \*
* The guy, who built the experimental firmware wanted to use a greater delay, so the test was done with cli option ```-t 60```. 

I do not know, how to optimize the delay, since this should be done in a more realistic scenario. I don't want to deploy the patch again and again to all nodes in my community.

-

\* Maybe some of the remaining loss is caused by this environment. If the random delay (of the patched router) is very short, the response may still conflict with the bulk of responses of the unpatched ones.

net/respondd/src/respondd.c
lemoer: No. The packet header information is stored in this buffer. The ```CMSG_*```-macros are used to access them, since different headers have different length. ```CMSG (3)```

lemoer:
@jplitza Pong and thank you for your feedback!

net/respondd/src/respondd.c
jplitza: Ah, I see, sorry.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+	int64_t timeout = schedule_idle_time(schedule);
+	if (timeout < 0)
+		return;
+
+	// set timeout to the socket
+	struct timeval t;
+	t.tv_sec = ((uint64_t) timeout) / 1000;
+	t.tv_usec = (((uint64_t) timeout) % 1000) * 1000;
+
+	if (setsockopt (sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&t, sizeof(t)) < 0)
```
NeoRaider: The `(char *)` cast is not required, setsockopt takes a `void *`. Also, remove the space after the function name to follow the coding style of the rest of the file.

net/respondd/src/respondd.c
```diff
@@ -165,6 +181,56 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+
+int64_t schedule_idle_time(struct request_schedule *s) {
+	if (!s->list_head)
+		// nothing to do yet (0 = infinite time)
+		return 0;
+
+	update_time();
+	int64_t result = s->list_head->scheduled_time - now;
+
+	if (result == 0)
```
NeoRaider: Maybe `if (result <= 0)`? Doesn't really matter though, but somehow feels more consistent.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+		break;
+	}
+
 	input[input_bytes] = 0;
 
-	struct json_object *result = handle_request(input, &compress);
-	if (!result)
-		return;
+	// only multicast requests are delayed
+	update_time();
```
NeoRaider: This `update_time()` should be moved directly below the `recvmsg` (before the error checks) to avoid skipping it on errors.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+	}
+
+	struct request_task *new_task = malloc(sizeof(struct request_task));
+	new_task->scheduled_time = now + delay;
+	strcpy(new_task->request, input);
+	memcpy(&new_task->client_addr, &addr, addrlen);
+
+	if(!schedule_push_request(schedule, new_task)) {
+		free(new_task);
+	}
```
NeoRaider: More nitpicks: please add a space after `if`, the braces can be removed.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+	int64_t delay;
+	if (IN6_IS_ADDR_MULTICAST(&destaddr)) {
+		delay = rand() % max_multicast_delay;
+	} else {
+		delay = 0;
+	}
+
+	struct request_task *new_task = malloc(sizeof(struct request_task));
+	new_task->scheduled_time = now + delay;
+	strcpy(new_task->request, input);
```
NeoRaider: Please use `strncpy` here and ensure `new_task->request` gets zero-terminated, otherwise this might break due to future changes...

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+	memcpy(&new_task->client_addr, &addr, addrlen);
```
NeoRaider: Both `new_task->client_addr` and `addr` are sockaddr_in6, you can just `new_task->client_addr = addr`.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+	// only multicast requests are delayed
+	update_time();
+	int64_t delay;
+	if (IN6_IS_ADDR_MULTICAST(&destaddr)) {
+		delay = rand() % max_multicast_delay;
+	} else {
+		delay = 0;
+	}
+
+	struct request_task *new_task = malloc(sizeof(struct request_task));
```
NeoRaider: I think `malloc(sizeof(*new_task))` is nicer.

net/respondd/src/respondd.c
```diff
@@ -64,6 +67,18 @@ struct request_type {
...
+struct request_task {
+	struct request_task *next;
+	int64_t scheduled_time;
+
+	struct sockaddr_in6 client_addr;
+	char request[REQUEST_MAXLEN];
+};
+
+struct request_schedule {
+	int length;
```
NeoRaider: I usually prefer `size_t` for lenghts and counts of any kind.

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+		delay = 0;
+	}
+
+	struct request_task *new_task = malloc(sizeof(struct request_task));
+	new_task->scheduled_time = now + delay;
+	strcpy(new_task->request, input);
+	memcpy(&new_task->client_addr, &addr, addrlen);
+
+	if(!schedule_push_request(schedule, new_task)) {
+		free(new_task);
```
NeoRaider: I don't like the fact that a full schedule can make respondd drop requests - and it is even possible to break unicast requests my sending to many multicasts!

Let's rather:
* Not queue requests with delay == 0 at all, but just serve them directly
* Serve requests instead of dropping them when the queueing fails

net/respondd/src/respondd.c
```diff
@@ -320,27 +384,99 @@ static struct json_object * handle_request(char *request, bool *compress) {
...
+		.msg_iovlen = 1,
+		.msg_control = control,
+		.msg_controllen = sizeof(control)
+	};
 
-	input_bytes = recvfrom(sock, input, sizeof(input)-1, 0, (struct sockaddr *)&addr, &addrlen);
+	input_bytes = recvmsg(sock, &mh, 0);
+
+	// Timeout
+	if (input_bytes < 0 && errno == EWOULDBLOCK)
```
NeoRaider: Let's change the errno check to `(errno == EAGAIN || errno == EWOULDBLOCK)`. They are the same on Linux, but it's better to make this portable in case someone wants to port respondd to other systems.

NeoRaider:
This is looking very good, I only found minor issues :)

lemoer:
@NeoRaider Thanks for the feedback. I tried to implement your comments.

>> I think 10 seconds default are a bit high. Do lower values, like maybe 5 seconds, yield similar results?
>
> I'm not sure about the best value for the maximum delay. The current test does not say anything about the behavior in the real scenario.
>
> Some more facts about the graph:
> - I have deployed the patch to a few routers (experimental branch) in the freifunk hanover community and the remaining routers are still answering to the multicast request immediately. \*
> - The guy, who built the experimental firmware wanted to use a greater delay, so the test was done with cli option -t 60.
>
> I do not know, how to optimize the delay, since this should be done in a more realistic scenario. I don't want to deploy the patch again and again to all nodes in my community.
>
>-
> \* Maybe some of the remaining loss is caused by this environment. If the random delay (of the patched router) is very short, the response may still conflict with the bulk of responses of the unpatched ones.

What do you think about this?

NeoRaider:
I think 10 seconds should be ok for now.

Besides the storm of the replies themselves, another thing that might overload the network are the neighbour solicitations for the querier. The first request after some time without requests will cause one multicast query from each node (!) at the moment, as the nodes need to resolve the querier IP to its MAC address. The batman-adv multicast optimizations that are currently being worked on will hopefully reduce the number of packets significantly.

I also have a few ideas for a "partitioned query" mode in which only a subset of the nodes answers a query. Example: Calculate a subset ID as a hash of the primary MAC modulo 16; now a partitioned query could be sent that is only answered by nodes with a specific subset ID. The querier can send one request for each of the 16 possible IDs to cover all nodes (the modulus could also be chosen by the querier). This could be a future extension to the respondd protocol.

On the other hand, mesh-wide respondd won't work with layer 3 routing protocols anyways, as we won't have multicast routing (at least in the beginning). We'll need to think of another way to discover the nodes then, so the partitioned query idea might be moot...

lemoer:
On 09/10/2016 04:57 AM, Matthias Schiffer wrote:
> Besides the storm of the replies themselves, another thing that might overload the network are the neighbour solicitations for the querier. The first request after some time without requests will cause one multicast query from each node (!) at the moment, as the nodes need to resolve the querier IP to its MAC address. The batman-adv multicast optimizations that are currently being worked on will hopefully reduce the number of packets significantly.

Ah. That's interesting. I have not thought about that yet.

> I also have a few ideas for a "partitioned query" mode in which only a subset of the nodes answers a query. Example: Calculate a subset ID as a hash of the primary MAC modulo 16; now a partitioned query could be sent that is only answered by nodes with a specific subset ID. The querier can send one request for each of the 16 possible IDs to cover all nodes (the modulus could also be chosen by the querier). This could be a future extension to the respondd protocol.

I also thougth about a similar (but a little bit simpler) approach,
which is already possible without protocol changes. The respondd daemon
on the freifunk router could listen on an additional multicast group
(maybe hash based on primary mac). So the network could be splitted in
some groups.

But I think this wont be necessary for now.

net/respondd/src/respondd.c
```diff
@@ -165,6 +181,56 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+	s->length++;
+	return true;
+}
+
+int64_t schedule_idle_time(struct request_schedule *s) {
+	if (!s->list_head)
+		// nothing to do yet (0 = infinite time)
+		return 0;
+
+	update_time();
```
NeoRaider: I don't think the `update_time()` here is necessary. The only part of respondd that will ever block is the recvmsg, after which the time is updated.

net/respondd/src/respondd.c
```diff
@@ -361,13 +408,118 @@ static void serve(int sock) {
...
+// 2b. If the schedule is full, send the reply immediately.
+// 2c. If the incomming request was sent to a unicast destination, the response
+//     will be also sent immediately.
+static void accept_request(struct request_schedule *schedule, int sock,
+                           uint64_t max_multicast_delay) {
+	char input[REQUEST_MAXLEN];
+	ssize_t input_bytes;
+	struct sockaddr_in6 addr;
+	char control[256];
+	struct in6_addr destaddr;
```
NeoRaider: Let's initialize destaddr here (`struct in6_addr destaddr = {};`), just to be sure. I seem to remember there might be messages without ip6_pktinfo (using v4-mapped addresses?).

net/respondd/src/respondd.c
```diff
@@ -361,13 +408,118 @@ static void serve(int sock) {
...
+		free(new_task);
+		return;
+	}
+
+	if (!schedule_push_request(schedule, new_task)) {
+		// we can't schedule, so send the response immediately
+		serve_request(new_task, sock);
+		free(new_task);
+	}
+}
```
NeoRaider: I don't like the code duplication here. Maybe use something like:
```
bool delayed = false;
if (IN6_IS_ADDR_MULTICAST(&destaddr))
  delayed = schedule_push_request(schedule, new_task);
if (delayed)
  return;
// Serve immediately here...
```

net/respondd/src/respondd.c
```diff
@@ -443,8 +612,19 @@ int main(int argc, char **argv) {
...
 	load_providers();
 
-	while (true)
-		serve(sock);
+	struct request_schedule schedule = {};
+
+	while (true) {
+		accept_request(&schedule, sock, max_multicast_delay);
+
+		struct request_task* task = schedule_pop_request(&schedule);
```
NeoRaider: Coding style: `struct request_task *task`

net/respondd/src/respondd.c
```diff
@@ -361,13 +408,118 @@ static void serve(int sock) {
...
+
+	send_response(
+		sock,
+		result,
+		compress,
+		&task->client_addr
+	);
+}
+
+// Wait for an incomming request and schedule it.
```
NeoRaider: Typo: "incoming"

NeoRaider:
Thanks for the improvements, and sorry for overlooking some minor issues in my first review.

Two more things:
* I generally prefer `/* */` for comments (you even used `//` for a quite long comment...)
* If you want, you can add your name to the license header.

lemoer:
@NeoRaider Pong :)

NeoRaider:
I've just noticed a severe issue of this PR: the status page uses respondd as well to get information from direct neighbours; in this usecase, the replies must not be delayed.

I see two possible options:
1. Add a delay field to the requests themselves (this could be done backwards- and forwards-compatible: terminate the request string with a 0 byte, and add the maximum delay afterwards (maybe using simple TLV records to keep it extensible?)
2. Mark certain interfaces as delayable (so requests on br-client would be delayed, but no requests on other interfaces) by adding an alternative command line option (keep `-i` for normal interfaces, use `-d` for delayable interfaces)

I don't like either of this options, but I currently tend towards the former.

jplitza:
I really dislike the first option, because the "protocol" is a hacky mess already, and it's planned to replace the whole protocol anyway, right? So I'm really against some ad-hoc changes to the protocol now. Even more so, because if we look at things from a DDoS side of things, making the delay optional and requester-configurable isn't a good idea.

But how much of a problem is this, really? The status page itself (as in, the Javascript application) is asynchronous anyway, so shouldn't a delayed response work just fine?

lemoer:
*CoAP* will replace our current protocol somewhen in future (Issue: freifunk-gluon/gluon#522).

I recently read the *CoAP* specification ```rfc 7252```. The section 8.2 specifies the response behavior in multicast scenarios. The protocol takes an approach quite similar to our ideas. The following calculation is used to calculate the maximum response delay:

```
lb_Leisure = S * G / R

S: size of packet to send
G: number of nodes in the multicast group
R: max desired data rate at the receiving node
```

This calculation is very interesting. It replaces our problem of different delays by another problem of finding out ```G``` for each interface.  Yet I'm not really sure how complex the new problem is.

I think we should really invest in the new protocol instead of extending the old protocol. What do you think?

(I am very interested in doing this, but I think the earliest time I could do this would be around the congress.)

lemoer:
I have implemented the interface based solution as hotfix for the problem. I decided to leave the protocol as it is.

If you want to use the patched version of respondd, you have to apply this patch to gluon:
```
diff --git a/package/gluon-respondd/files/etc/init.d/gluon-respondd b/package/gluon-respondd/files/etc/init.d/gluon-respondd
index a1bf979..a35ddb3 100755
--- a/package/gluon-respondd/files/etc/init.d/gluon-respondd
+++ b/package/gluon-respondd/files/etc/init.d/gluon-respondd
@@ -13,7 +13,7 @@ LOCK=/var/run/gluon-respondd.lock
 
 
 do_start() {
-	DEVS=$(cat $DEVLIST 2>/dev/null | while read dev iface; do echo -n " -i $dev"; done)
+	DEVS=$(cat $DEVLIST 2>/dev/null | while read dev iface; do echo -n " -i $dev"; if [ "$dev" != "br-client" ]; then echo -n " -t 0"; fi; done)
 	service_start $DAEMON -g ff02::2:1001 -p 1001 -d /lib/gluon/respondd $DEVS
 }
 ```

I'm not sure, whether we should merge this upstream? We are using it now in Freifunk Hanover.

File: net/respondd/src/respondd.c [hunk 26]
```diff
@@ -416,9 +602,33 @@ int main(int argc, char **argv) {
...
 			join_mcast(sock, mgroup_addr, optarg);
 			break;
 
+		case 't':
+			if (!iface_set) {
+				fprintf(stderr, "Interface must be given before max response delay.\n");
+				exit(EXIT_FAILURE);
+			}
+
+			uint64_t max_multicast_delay = 1000 * strtoul(optarg, &endptr, 10);
```
NeoRaider: You must cast one of the factors to `uint64_t`, otherwise they will be multiplied as `unsigned long` and you lose the additional range provided by the `uint64_t`. The nicest way to achieve this would be to write `UINT64_C(1000)` (this macro is defined in `stdint.h`).

net/respondd/src/respondd.c
```diff
@@ -361,13 +417,132 @@ static void serve(int sock) {
...
+		.msg_iovlen = 1,
+		.msg_control = control,
+		.msg_controllen = sizeof(control)
+	};
+
+	input_bytes = recvmsg(sock, &mh, 0);
+	update_time();
+
+	// Timeout
+	if (input_bytes < 0 && (errno == EAGAIN || errno == EWOULDBLOCK))
```
NeoRaider: `update_time()` may change `errno`, it must be saved in another variable.

net/respondd/src/respondd.c
```diff
@@ -361,13 +417,132 @@ static void serve(int sock) {
...
+	}
+
+	input[input_bytes] = 0;
+
+	// get the max delay
+	uint64_t max_multicast_delay = MAX_MULTICAST_DELAY_DEFAULT;
+	struct interface_delay_info *tmp = if_delay_info_list;
+	for (; tmp; tmp = tmp->next) {
+		if (tmp->ifindex == ifindex)
+			max_multicast_delay = tmp->max_multicast_delay;
```
NeoRaider: Also add a break here to skip the remaining loop iterations when an entry has been found.

net/respondd/src/respondd.c
```diff
@@ -361,13 +417,132 @@ static void serve(int sock) {
...
+ */
+static void accept_request(struct request_schedule *schedule, int sock,
+                           struct interface_delay_info *if_delay_info_list) {
+	char input[REQUEST_MAXLEN];
+	ssize_t input_bytes;
+	struct sockaddr_in6 addr;
+	char control[256];
+	struct in6_addr destaddr = {};
+	struct cmsghdr *cmsg;
+	unsigned int ifindex;
```
NeoRaider: Intialize `ifindex` with 0, the compiler is warning it may be used unitialized.

net/respondd/src/respondd.c
```diff
@@ -416,9 +602,33 @@ int main(int argc, char **argv) {
 				fprintf(stderr, "Multicast group must be given before interface.\n");
 				exit(EXIT_FAILURE);
 			}
+			iface_set = true;
+			last_ifindex = if_nametoindex(optarg);
```
lemoer: Implemented in d68f281.

net/respondd/src/respondd.c
```diff
@@ -361,13 +417,132 @@ static void serve(int sock) {
 	}
 
 	if (output) {
-		if (sendto(sock, output, output_bytes, 0, (struct sockaddr *)&addr, addrlen) < 0)
+		if (sendto(sock, output, output_bytes, 0, addr, sizeof(struct sockaddr_in6)) < 0)
```
lemoer: I understand the idea, but i can't reproduce the compiler warning. Neither with clang nor with gcc.

```
$CC -ldl -ljson-c -std=c99 -D_GNU_SOURCE miniz.c respondd.c -Wall -Wextra -pedantic
```

net/respondd/src/respondd.c
lemoer: Same as above.

NeoRaider: This one appears when `-O2` or `-Os` are passed.

net/respondd/src/respondd.c
NeoRaider: glibc uses the `__attribute__((__transparent_union__))` extension to make socket functions work with any sockaddr type, so no warning is shown. This is neither standard-compliant nor do other libcs like musl implement it.

lemoer:
@NeoRaider I hopefully implemented all of your requested changes.

NeoRaider:
Thanks, merged! I'll add the delay to the command line as soon as https://github.com/freifunk-gluon/gluon/pull/946 is finished.

NeoRaider:
The delay is now active in the Gluon master.

PR #146, opened 2016-09-08
batman-adv-legacy: ignore error when writing to mesh_iface

rotanid:
this should fix https://github.com/freifunk-gluon/gluon/issues/820


rotanid:
acked by @NeoRaider on IRC

PR #153, opened 2016-12-08
respondd: Accept multiple data provider directories

jplitza:
Intention is to have one directory for Gluon-specific providers and one for general ones.

*Edit*: Hm, this doesn't totally work yet. If we have multiple data provider directories, where should the cache times be defined?

### DIFF
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -34,6 +34,7 @@
 #include <alloca.h>
 #include <dirent.h>
 #include <dlfcn.h>
+#include <fcntl.h>
 #include <inttypes.h>
 #include <search.h>
 #include <stdbool.h>
@@ -49,6 +50,7 @@
 #include <netinet/in.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
 
 #define SCHEDULE_LEN 8
 #define REQUEST_MAXLEN 256
@@ -99,13 +101,13 @@ static struct json_object * merge_json(struct json_object *a, struct json_object
 static void usage() {
 	puts("Usage:");
 	puts("  respondd -h");
-	puts("  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir>]");
+	puts("  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir> [-d <dir> ..]]");
 	puts("        -p <int>         port number to listen on");
 	puts("        -g <ip6>         multicast group, e.g. ff02::2:1001");
 	puts("        -i <string>      interface on which the group is joined");
 	puts("        -t <int>         maximum delay seconds before multicast responses");
 	puts("                         for the last specified multicast interface (default: 0)");
-	puts("        -d <string>      data provider directory (default: current directory)");
+	puts("        -d <string>      data provider directory");
 	puts("        -h               this help\n");
 }
 
@@ -242,10 +244,6 @@ struct request_task * schedule_pop_request(struct request_schedule *s) {
 }
 
 static void load_cache_time(struct request_type *r, const char *name) {
-	r->cache = NULL;
-	r->cache_time = 0;
-	r->cache_timeout = now;
-
 	char filename[strlen(name) + 7];
 	snprintf(filename, sizeof(filename), "%s.cache", name);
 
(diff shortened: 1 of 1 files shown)


net/respondd/src/respondd.c
```diff
@@ -171,13 +171,9 @@ static struct json_object * merge_json(struct json_object *a, struct json_object
...
 
-static const struct respondd_provider_info * get_providers(const char *filename) {
-	/*
-	  Prefix the filename with "./" to open the module in the current directory
-	  (dlopen looks in the standard library paths by default)
-	*/
-	char path[2 + strlen(filename) + 1];
-	snprintf(path, sizeof(path), "./%s", filename);
+static const struct respondd_provider_info * get_providers(const char *path, const char *filename) {
+	char path[strlen(path) + strlen(filename) + 2];
```
NeoRaider: You're shadowing the argument `path` here.

net/respondd/src/respondd.c
```diff
@@ -99,13 +99,13 @@ static struct json_object * merge_json(struct json_object *a, struct json_object
 static void usage() {
 	puts("Usage:");
 	puts("  respondd -h");
-	puts("  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir>]");
+	puts("  respondd [-p <port>] [-g <group> -i <if0> [-i <if1> ..]] [-d <dir> [-d <dir> ..]");
```
NeoRaider: Unbalanced brackets, there is one `]` missing.

net/respondd/src/respondd.c
```diff
@@ -292,16 +288,10 @@ static void add_provider(const char *name, const struct respondd_provider_info *
...
-	if (!hcreate_r(32, &htab)) {
-		perror("hcreate_r");
-		exit(EXIT_FAILURE);
-	}
-
-	DIR *dir = opendir(".");
+	DIR *dir = opendir(path);
 	if (!dir) {
 		perror("opendir");
 		exit(EXIT_FAILURE);
```
NeoRaider: I think it might be a good idea to ignore failures here, so we can unconditionally add certain dirs to the command line, even when no providers are installed into the dir.

net/respondd/src/respondd.c
```diff
@@ -649,10 +645,7 @@ int main(int argc, char **argv) {
 			break;
 
 		case 'd':
-			if (chdir(optarg)) {
```
NeoRaider: Without the `chdir`, we will to define where to store cache files in a different way.

jplitza: Yes, I just noticed that too.

Easiest fix would be to `chdir` to every directory that is given inside the `load_providers` function. That way, the first directory that has providers for a given request type has the chance to register the cache time.

Would that be okay?

NeoRaider: Seems a bit arbitrary to me. At least, every directory should have a chance to provide a cache time.

jplitza: So which would have precedence?

NeoRaider: Doesn't really matter if it's the first or the last, whatever is nicer to implement...

jplitza: Now I open the cache file for every provider that is loaded, thus also in every directory a request type is defined in, and the last one has precedence because earlier values are simply overridden if the file exists.

net/respondd/src/respondd.c
```diff
@@ -292,20 +288,15 @@ static void add_provider(const char *name, const struct respondd_provider_info *
...
-	}
+	if (!chdir(path))
+		return;
 
-	DIR *dir = opendir(".");
-	if (!dir) {
-		perror("opendir");
-		exit(EXIT_FAILURE);
-	}
+	DIR *dir = opendir(path);
```
NeoRaider: This will fail now if `path` is relative.

jplitza: Yes, sorry, forgot to change this back after re-introducing the chdir. Fixed.

net/respondd/src/respondd.c
```diff
@@ -292,20 +288,15 @@ static void add_provider(const char *name, const struct respondd_provider_info *
...
-static void load_providers(void) {
+static void load_providers(const char *path) {
 	update_time();
 
-	/* Maximum number of request types, might be made configurable in the future */
-	if (!hcreate_r(32, &htab)) {
-		perror("hcreate_r");
-		exit(EXIT_FAILURE);
-	}
+	if (!chdir(path))
```
NeoRaider: You still need to chdir back at some point, otherwise relative paths not given as first argument will be interpreted incorrectly.

net/respondd/src/respondd.c
```diff
@@ -292,19 +290,20 @@ static void add_provider(const char *name, const struct respondd_provider_info *
...
 
-static void load_providers(void) {
+static void load_providers(const char *path) {
 	update_time();
 
-	/* Maximum number of request types, might be made configurable in the future */
-	if (!hcreate_r(32, &htab)) {
-		perror("hcreate_r");
-		exit(EXIT_FAILURE);
+	int cwdfd = open(".", O_DIRECTORY);
```
jplitza: I would have loved to open with `O_PATH` here, but I wasn't able to get the build environment to recognize the option, even when I included all headers and feature macros that the manpage mentioned.

PR #156, opened 2016-12-16
Add package respondd-module-airtime

jplitza:
This supersedes freifunk-gluon/gluon#826

### DIFF
--- a/net/respondd-module-airtime/Makefile
+++ b/net/respondd-module-airtime/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=respondd-module-airtime
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_LICENSE:=BSD-2-Clause
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+PKG_BUILD_DEPENDS := respondd
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/respondd-module-airtime
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Add airtime to respondd
+  DEPENDS:=+respondd +libnl-tiny
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+	$(CP) ./src/* $(PKG_BUILD_DIR)/
+endef
+
+define Build/Configure
+endef
+
+
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include/libnl-tiny
+
+define Build/Compile
+	CFLAGS="$(TARGET_CFLAGS)" CPPFLAGS="$(TARGET_CPPFLAGS)" $(MAKE) -C $(PKG_BUILD_DIR) $(TARGET_CONFIGURE_OPTS)
+endef
+
+
+define Package/respondd-module-airtime/install
+	$(INSTALL_DIR) $(1)/lib/respondd
+	$(CP) $(PKG_BUILD_DIR)/respondd.so $(1)/lib/respondd/airtime.so
+endef
+
+$(eval $(call BuildPackage,respondd-module-airtime))
--- a/net/respondd-module-airtime/src/Makefile
+++ b/net/respondd-module-airtime/src/Makefile
@@ -13,4 +13,4 @@ respondd.so: airtime.c ifaces.c respondd.c
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -lnl-tiny -o $@ $^ $(LDLIBS)
 
 clean:
-	rm -rf *.so airtime-test
+	rm -rf *.so
--- a/net/respondd-module-airtime/src/Makefile
+++ b/net/respondd-module-airtime/src/Makefile
@@ -1,0 +1,16 @@
+# standard compliance
+CFLAGS         += -std=c99
+
+# warnings
+CFLAGS         += -Wall -Wextra -Wformat=2 -Wshadow -Wpointer-arith
+CFLAGS         += -pedantic
+
+all: respondd.so
+
+%.c: %.h
+
+respondd.so: airtime.c ifaces.c respondd.c
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -lnl-tiny -o $@ $^ $(LDLIBS)
+
+clean:
+	rm -rf *.so airtime-test
--- a/net/respondd-module-airtime/src/airtime.c
+++ b/net/respondd-module-airtime/src/airtime.c
@@ -100,13 +100,13 @@ abort:
 }
 
 bool get_airtime(struct airtime_result *result, int ifx) {
-	bool ok;
+	bool ok = false;
 	int ctrl;
 	struct nl_sock *sk = NULL;
 	struct nl_msg *msg = NULL;
 
 
-#define CHECK(x) { if (!(x)) { fprintf(stderr, "%s: error on line %d\n", __FILE__, __LINE__); ok = false; goto out; } }
+#define CHECK(x) { if (!(x)) { fprintf(stderr, "%s: error on line %d\n", __FILE__, __LINE__); goto out; } }
 
 	CHECK(sk = nl_socket_alloc());
 	CHECK(genl_connect(sk) >= 0);
--- a/net/respondd-module-airtime/src/airtime.c
+++ b/net/respondd-module-airtime/src/airtime.c
@@ -78,13 +78,13 @@ static int survey_airtime_handler(struct nl_msg *msg, void *arg) {
 		goto abort;
 	}
 
-	if(nla_parse_nested(sinfo, NL80211_SURVEY_INFO_MAX, tb[NL80211_ATTR_SURVEY_INFO], survey_policy)) {
+	if (nla_parse_nested(sinfo, NL80211_SURVEY_INFO_MAX, tb[NL80211_ATTR_SURVEY_INFO], survey_policy)) {
 		fprintf(stderr, "failed to parse nested attributes!\n");
 		goto abort;
 	}
 
 	// Channel active?
-	if(!sinfo[NL80211_SURVEY_INFO_IN_USE]){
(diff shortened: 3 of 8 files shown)


net/respondd-module-airtime/src/ifaces.c
```diff
@@ -0,0 +1,64 @@
...
+#include <linux/nl80211.h>
+#include <netlink/netlink.h>
+#include <netlink/genl/genl.h>
+#include <netlink/genl/ctrl.h>
+#include <net/if.h>
+#include <stdlib.h>
+
+#include "ifaces.h"
+
+static int iface_dump_handler(struct nl_msg *msg, struct iface_list **arg) {
```
NeoRaider: Please change the second parameter type to `void *` and remove the *nl_recvmsg_msg_cb_t* cast below. ISO C doesn't generally guarantee that all pointers have the same size, alignment or representation (only that any pointer can be converted to `void *` and back), so doing a function pointer cast like the one below might not be entirely correct.

(Note that this is only a cosmetic issue in practice; I think POSIX will guarantee that data pointers have the same representation. I'm not entirely sure though.)

net/respondd-module-airtime/src/respondd.c
```diff
@@ -0,0 +1,58 @@
...
+	json_object_object_add(obj, "noise",     json_object_new_int(air->noise));
+
+	json_object_array_add(wireless, obj);
+}
+
+static struct json_object *respondd_provider_statistics(void) {
+	struct airtime_result airtime = {};
+	struct json_object *result, *wireless;
+	struct iface_list *ifaces;
+	void *freeptr;
```
NeoRaider: Please move the definition of *freeptr* into the while loop below to make its scope clearer.

net/respondd-module-airtime/src/airtime.c
```diff
@@ -0,0 +1,136 @@
...
+	result->rx_time     = nla_get_u64(sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_RX]);
+	result->tx_time     = nla_get_u64(sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_TX]);
+	result->noise       = nla_get_u8(sinfo[NL80211_SURVEY_INFO_NOISE]);
+
+abort:
+	return NL_SKIP;
+}
+
+int get_airtime(struct airtime_result *result, int ifx) {
+	int error = 0;
```
NeoRaider: No error codes except for 0 and 1 are used, please change it to *bool* and make the return true by default.

net/respondd-module-airtime/src/airtime.c
```diff
@@ -0,0 +1,136 @@
...
+	genlmsg_put(msg, 0, 0, ctrl, 0, NLM_F_DUMP, NL80211_CMD_GET_SURVEY, 0);
+
+	NLA_PUT_U32(msg, NL80211_ATTR_IFINDEX, ifx);
+
+	CHECK(nl_send_auto_complete(sk, msg) >= 0);
+	CHECK(nl_recvmsgs_default(sk) >= 0);
+
+#undef CHECK
+
+nla_put_failure:
```
NeoRaider: I think it would make sense to set *error* between these two labels?

net/respondd-module-airtime/src/airtime.c
```diff
@@ -0,0 +1,136 @@
...
+		goto abort;
+	}
+
+	if(nla_parse_nested(sinfo, NL80211_SURVEY_INFO_MAX, tb[NL80211_ATTR_SURVEY_INFO], survey_policy)) {
+		fprintf(stderr, "failed to parse nested attributes!\n");
+		goto abort;
+	}
+
+	// Channel active?
+	if(!sinfo[NL80211_SURVEY_INFO_IN_USE]){
```
NeoRaider: The whilespace is inconsistent in the if expressions in this function. Please add a space before `(` and after `)` in l.81 and l.87.

net/respondd-module-airtime/src/respondd.c
```diff
@@ -0,0 +1,58 @@
...
+
+	wireless = json_object_new_array();
+	if (!wireless) {
+		json_object_put(result);
+		return NULL;
+	}
+
+	ifaces = get_ifaces();
+	while (ifaces != NULL) {
+		get_airtime(&airtime, ifaces->ifx);
```
NeoRaider: Please check the return value of *get_airtime()*.

net/respondd-module-airtime/src/ifaces.c
```diff
@@ -0,0 +1,64 @@
...
+#define CHECK(x) { if (!(x)) { fprintf(stderr, "airtime.c: error on line %d\n",  __LINE__); goto out; } }
+
+	CHECK(sk = nl_socket_alloc());
+	CHECK(genl_connect(sk) >= 0);
+
+	CHECK(ctrl = genl_ctrl_resolve(sk, NL80211_GENL_NAME));
+	CHECK(nl_socket_modify_cb(sk, NL_CB_VALID, NL_CB_CUSTOM, (nl_recvmsg_msg_cb_t) iface_dump_handler, &ifaces) == 0);
+	CHECK(msg = nlmsg_alloc());
+
+	/* TODO: check return? */
```
NeoRaider: Yes.

net/respondd-module-airtime/src/ifaces.c
```diff
@@ -0,0 +1,64 @@
...
+	for (last_next = arg; *last_next != NULL; last_next = &(*last_next)->next) {
+		if ((*last_next)->wiphy == wiphy)
+			goto abort;
+	}
+	*last_next = malloc(sizeof(**last_next));
+	(*last_next)->next = NULL;
+	(*last_next)->ifx = nla_get_u32(tb[NL80211_ATTR_IFINDEX]);
+	(*last_next)->wiphy = wiphy;
+
+abort:
```
NeoRaider: Maybe call this label `skip`?

net/respondd-module-airtime/src/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+all: respondd.so
+
+%.c: %.h
+
+respondd.so: airtime.c ifaces.c respondd.c
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -shared -fPIC -D_GNU_SOURCE -lnl-tiny -o $@ $^ $(LDLIBS)
+
+clean:
+	rm -rf *.so airtime-test
```
NeoRaider: `airtime-test` here is obsolete.

net/respondd-module-airtime/src/respondd.c
```diff
@@ -0,0 +1,58 @@
+#include <string.h>
+#include <stdio.h>
+#include <json-c/json.h>
+#include <respondd.h>
+
+#include "airtime.h"
+#include "ifaces.h"
+
+void fill_airtime_json(struct airtime_result *air, struct json_object *wireless) {
```
NeoRaider: This function should be static.

net/respondd-module-airtime/src/ifaces.c
```diff
@@ -0,0 +1,64 @@
...
+	return NL_SKIP;
+}
+
+struct iface_list *get_ifaces() {
```
NeoRaider: I'm not happy with these publicly visible functions in a dynamically linked module. Please add `__attribute__ ((visibility ("hidden")))`. (Also for *get_airtime()*)

net/respondd-module-airtime/Makefile
```diff
@@ -0,0 +1,42 @@
...
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include/libnl-tiny
+
+define Build/Compile
+	CFLAGS="$(TARGET_CFLAGS)" CPPFLAGS="$(TARGET_CPPFLAGS)" $(MAKE) -C $(PKG_BUILD_DIR) $(TARGET_CONFIGURE_OPTS)
+endef
+
+
+define Package/respondd-module-airtime/install
+	$(INSTALL_DIR) $(1)/lib/respondd
+	$(CP) $(PKG_BUILD_DIR)/respondd.so $(1)/lib/respondd/airtime.so
```
corny: Do we use `/lib/respondd` or `/lib/gluon/respondd`? With Gluon 2016.2.2 only the last version works.

jplitza: The package is not intended to be backported, in favor of being independent from Gluon. So we are using the first, generic path.

corny:
Successfully tested on:
* WDR3600
* WDR4900
* CPE210

net/respondd-module-airtime/src/airtime.c
NeoRaider: This still needs to be done.

net/respondd-module-airtime/src/respondd.c
```diff
@@ -0,0 +1,58 @@
...
+
+	wireless = json_object_new_array();
+	if (!wireless) {
+		json_object_put(result);
+		return NULL;
+	}
+
+	ifaces = get_ifaces();
+	while (ifaces != NULL) {
+		if(get_airtime(&airtime, ifaces->ifx))
```
NeoRaider: Formatting: `if (`

net/respondd-module-airtime/src/airtime.c
NeoRaider: Hmm okay, just setting *ok* here won't work, we'd also need to add a `goto out` just above this label.

corny: done

net/respondd-module-airtime/src/airtime.c
```diff
@@ -0,0 +1,137 @@
...
+	result->rx_time     = nla_get_u64(sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_RX]);
+	result->tx_time     = nla_get_u64(sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_TX]);
+	result->noise       = nla_get_u8(sinfo[NL80211_SURVEY_INFO_NOISE]);
+
+abort:
+	return NL_SKIP;
+}
+
+bool get_airtime(struct airtime_result *result, int ifx) {
+	bool ok;
```
NeoRaider: You must initialize *ok* with false, it won't be set in the *nla_put_failure* path otherwise.

The initlalization in the *CHECK* error path can be removed then.

corny: Sorry, you're absolutely right. Is fixed now.

PR #161, opened 2017-02-13
net/mmfd: add experimental mmfd package

christf:
this adds the experimental mmfd package that is used in the current babel-PR.


christf:
let's let this sit until it was decided whether mmfd might get a new home under the gluon umbrella. also mmfd has been patched. let's include these in the first version of the package as well.

christf:
patches and new package git home included.

net/mmfd/files/lib/gluon/upgrade/430-add-mmfd-interface
```diff
@@ -0,0 +1,12 @@
...
+local uci = require('luci.model.uci').cursor()
+
+uci:delete('network', 'mmfd')
+uci:section('network', 'interface', 'mmfd', {
+	proto = 'static',
+	ifname = 'mmfd0',
+	ip6addr = 'fe80::1/64'
+}
+)
+uci:save('network')
```
NeoRaider: This file is Gluon-specific and must be move into the *gluon-mesh-babel* package. Also, it still uses the old luci.model package.

NeoRaider: I think it would be nice if mmfd would just setup its address itself, without relying on netifd.

PR #164, opened 2017-02-27
autoupdater: new implementation

jplitza:
Implementation in C.

This is not yet well tested (but at least a bit).


admin/autoupdater/Makefile
```diff
@@ -2,30 +2,32 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=autoupdater
 PKG_VERSION:=2
+PKG_BUILD_DEPENDS:=libuclient libecdsautil libplatforminfo libuci
```
NeoRaider: Runtime depends don't need to be duplicated as build depends.

admin/autoupdater/src/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+project(AUTOUPDATER C)
+
+set_property(DIRECTORY PROPERTY COMPILE_DEFINITIONS _GNU_SOURCE)
+
+find_path(UBOX_INCLUDE_DIR NAMES libubox/uloop.h)
+find_library(UBOX_LIBRARY NAMES ubox)
+find_library(UCLIENT_LIBRARY NAMES uclient)
+
+find_library(UCI_LIBRARY NAMES uci)
+find_library(PLATFORMINFO_LIBRARY NAMES platforminfo)
```
NeoRaider: It would be better to write proper CMake package scripts for these libraries instead of using *find_library* directly. See https://gitlab.com/neoraider/ece/tree/master/cmake for a few examples.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
+/*
+  Copyright (c) 2016, Matthias Schiffer <mschiffer@universe-factory.net>
```
NeoRaider: It's 2017 now :P

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
...
+
+#include <sys/types.h>
+#include <sys/file.h>
+#include <sys/stat.h>
+
+#include "manifest.h"
+#include "settings.h"
+#include "uclient.h"
+#include "util.h"
+#include "version.h"
```
NeoRaider: It is usually a good idea to order includes from most specific (project-internal headers) to least specific (C standard, POSIX), to make sure that project-internal headers don't miss any includes.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
...
+		/*
+		 When the difference is negative, there are two possibilities: the
+		 manifest contains an incorrect date, or our own clock is wrong. As there
+		 isn't anything sensible to do for an incorrect manifest, we'll assume
+		 the latter is the case and update anyways as we can't do anything better
+		*/
+		fputs("autoupdater: warning: clock seems to be incorrect.\n", stderr);
+
+		if (get_uptime() < 600)
+			/* If the uptime is very low, it's possible we just didn't get the time over NTP yet, so we'll just wait until the next time the updater runs */
```
NeoRaider: Very long comment.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
...
+
+
+/** Receives data from uclient, chops it to lines and hands it to \ref parse_line */
+static void recv_manifest_cb(struct uclient *cl) {
+	static char buf[MAX_LINE_LENGTH + 1];
+	static char *ptr = buf;
+	char *newline;
+	int len;
+
+	while (1) {
```
NeoRaider: I think it's nicer to write `while (true)` instead.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
...
+		if (ptr - buf == MAX_LINE_LENGTH) {
+			fputs("autoupdater: error: encountered manifest line exceeding limit of " STRINGIFY(MAX_LINE_LENGTH) " characters\n", stderr);
+			break;
+		}
+		len = uclient_read(cl, ptr, MAX_LINE_LENGTH - (ptr - buf));
+		if (!len)
```
NeoRaider: You must also check for negative error return codes.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,411 @@
...
+	free_manifest(&G.manifest);
+	memset(&G.manifest, 0, sizeof(G.manifest));
+
+	/* Construct manifest URL */
+	char manifest_url[strlen(mirror) + strlen(G.settings.branch) + 11];
+	sprintf(manifest_url, "%s/%s.manifest", mirror, G.settings.branch);
+
+	/* Download manifest */
+	ecdsa_sha256_init(&G.manifest.hash_ctx);
+	get_url(manifest_url, recv_manifest_cb);
```
NeoRaider: The download functions definitely need better error handling. Obviously the autoupdater will error out as soon as the checksums are checked, but "download failed" would be a much more useful error message.

admin/autoupdater/src/hexutil.c
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+#include "hexutil.h"
+
+#include <stdio.h>
+#include <string.h>
+
+int parsehex(void *buffer, const char *string, size_t len) {
+  // number of digits must be even
+  if ((strlen(string) & 1) == 1)
```
NeoRaider: This file uses spaces for indentation, unlike the others. Please convert this to tabs, too.

admin/autoupdater/src/hexutil.c
```diff
@@ -0,0 +1,58 @@
...
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#include "hexutil.h"
+
+#include <stdio.h>
+#include <string.h>
+
+int parsehex(void *buffer, const char *string, size_t len) {
```
NeoRaider: *parsehex* uses a `void *` buffer, *dumphex* has `unsigned char *`. For the sake of consistency, please make this `unsigned char *`, too.

admin/autoupdater/src/settings.c
```diff
@@ -0,0 +1,178 @@
...
+
+	size_t i = 0;
+	struct uci_element *e;
+	uci_foreach_element(&o->v.list, e)
+		i++;
+
+	*len = i;
+	const char **ret = malloc(i * sizeof(char *));
+
+	o = uci_lookup_option(ctx, s, option);
```
NeoRaider: It setting *o* here again really necessary?

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,114 @@
...
+
+
+static void request_done(struct uclient *cl) {
+	uclient_disconnect(cl);
+	uloop_end();
+}
+
+
+static void header_done_cb(struct uclient *cl) {
+	static int retries;
```
NeoRaider: Using a static variable here that is shared by all requests seems weird. You could use the *priv* field of the *uclient* struct to point to a struct with such variables; the same struct could be used to signal error conditions back to *get_url*.

admin/autoupdater/src/CMakeLists.txt
jplitza: Can you explain why? I thought about this, but it seemed to only clutter the tree with files without having any advantages.

admin/autoupdater/src/hexutil.c
jplitza: I think the other way round would be more fitting, because it actually can be any data structure we are putting the binary data in/reading it from.

And by the way, the whole file was copied verbatim from ecdsautils. Of course I can make these changes, but maybe encourage @tcatm to also make them there. ;-)

jplitza:
Wow, that was fast! And on top, I was able to learn some better C style. Thanks!
I fixed some of the things you mentioned, but thanks to your hint with the `priv` field of uclient, I will probably restructure quite a bit to get rid of global variables and improve error handling.

Also, I think adding at least some basic tests for parsing the manifest would be a good idea. So don't rush to a more thorough review, you gave me enough to work with already. 😉

admin/autoupdater/src/autoupdater.c
jplitza: Github didn't notice, but I fixed this. Also, thanks for the explanation, that was new to me but makes perfect sense!

jplitza:
Okay, I was too enthusiastic to put this away, so I've made the restructuring I had in mind. Feel free to comment again! ☺

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,147 @@
...
+
+struct uclient_data {
+	/* data that can be passed in by caller and used in custom callbacks */
+	void *custom;
+	/* data used by uclient callbacks */
+	int retries;
+	int err_code;
+};
+
+#define uc_data(cl) ((struct uclient_data *)cl->priv)
```
NeoRaider: Using a static inline function would be cleaner than a macro.

admin/autoupdater/src/uclient.h
```diff
@@ -0,0 +1,33 @@
...
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#pragma once
+
+
+#include <libubox/uclient.h>
+
+#define uclient_get_custom(cl) *(void **)cl->priv
```
NeoRaider: Relying on the internal structure of a struct with pointer magic, instead of just pulling out the struct definitions to the header file, doesn't seem like a good idea.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,482 @@
...
+
+#define MAX_LINE_LENGTH 512
+#define STRINGIFY(str) #str
+
+const char *const download_d_dir = "/usr/lib/autoupdater/download.d";
+const char *const abort_d_dir = "/usr/lib/autoupdater/abort.d";
+const char *const upgrade_d_dir = "/usr/lib/autoupdater/upgrade.d";
+const char *const lockfile = "/var/run/autoupdater.lock";
+const char *const firmware_path = "/tmp/firmware.bin";
+const char *const sysupgrade_path = "/sbin/sysupgrade";
```
NeoRaider: These should also be static.

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,147 @@
...
+	static char http_code_errmsg[16];
+	if (code & UCLIENT_ERROR_STATUS_CODE) {
+		snprintf(http_code_errmsg, 16, "HTTP error %d",
+			code ^ UCLIENT_ERROR_STATUS_CODE);
+		return http_code_errmsg;
+	}
+	switch(code) {
+	case UCLIENT_ERROR_CONNECT:
+		return "Connection failed";
+		break;
```
NeoRaider: These breaks after returns can be removed.

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,147 @@
...
+	UCLIENT_ERROR_CONNECTION_RESET_PREMATURELY,
+	UCLIENT_ERROR_STATUS_CODE = 1024,
+};
+
+
+const char *uclient_get_errmsg(int code) {
+	static char http_code_errmsg[16];
+	if (code & UCLIENT_ERROR_STATUS_CODE) {
+		snprintf(http_code_errmsg, 16, "HTTP error %d",
+			code ^ UCLIENT_ERROR_STATUS_CODE);
```
NeoRaider: `code & (~UCLIENT_ERROR_STATUS_CODE)` would be more explicit for removing the UCLIENT_ERROR_STATUS_CODE bit.

admin/autoupdater/src/util.c
```diff
@@ -0,0 +1,99 @@
...
+
+	for (size_t i = 0; i < n_files; i++) {
+		pid_t pid = fork();
+		if (pid == 0) {
+			int null_fd = open("/dev/null", O_RDWR);
+			dup2(null_fd, 0);
+			dup2(null_fd, 1);
+			dup2(null_fd, 2);
+			for (int i = 3; i < 65536; i++)
+				close(i);
```
NeoRaider: I don't think this is necessary here. The autoupdater is single-threaded, and you don't have any uclient sessions still open when run_dir() is called.

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,147 @@
...
+
+#include <libubox/uloop.h>
+
+#include <limits.h>
+#include <stdio.h>
+
+
+#define TIMEOUT_MSEC 300000
+
+const char *const user_agent = "Gluon Autoupdater (using libuclient)";
```
NeoRaider: static

admin/autoupdater/src/settings.c
```diff
@@ -0,0 +1,178 @@
...
+		exit(1);
+	}
+
+	settings->good_signatures = load_positive_number(ctx, branch, "good_signatures");
+	if (settings->n_mirrors == 0)
+		settings->mirrors = load_string_list(ctx, branch, "mirror", &settings->n_mirrors);
+
+	const char **pubkeys_str = load_string_list(ctx, branch, "pubkey", &settings->n_pubkeys);
+	settings->pubkeys = malloc(settings->n_pubkeys * sizeof(ecc_25519_work_t));
+	int ignored_keys = 0;
```
NeoRaider: For consistency, make *ignored_keys* a size_t, too.

admin/autoupdater/src/manifest.h
```diff
@@ -0,0 +1,54 @@
...
+	char *image_filename;
+	unsigned char *image_hash[ECDSA_SHA256_HASH_SIZE];
+	char *version;
+	time_t date;
+	float priority;
+
+	size_t n_signatures;
+	ecdsa_signature_t **signatures;
+	ecdsa_sha256_context_t hash_ctx;
+};
```
NeoRaider: I think it would be nice to separate the actual manifest data from parser state like *sep_found*.

admin/autoupdater/src/util.c
jplitza: It definitely is: We've got the lockfile open! If we don't close that, the lock will be held by services resumed after a failed download, leading to the autoupdater not starting up again.

Now you can argue that we should somehow pass that specific fd into the function and only close that, but it kind of feels cleaner this way, IMO.

admin/autoupdater/src/manifest.h
jplitza: Well, what is `branch_ok`? We don't parse the whole manifest into variables, so I don't see any advantage in separating those.

admin/autoupdater/src/util.c
NeoRaider: Hmm. Just closing the lockfile seems cleaner to me, but I'm okay with your solution if you replace the arbitrary limit something like https://git.universe-factory.net/fastd/tree/src/fastd.c#n248 .

admin/autoupdater/src/manifest.h
NeoRaider: Okay, it probably doesn't really matter, and indeed *sep_found* is the only variable that's just parser state.

admin/autoupdater/src/util.c
```diff
@@ -26,73 +26,41 @@
...
-		strcpy(filenames[n_files], file->d_name);
-		n_files++;
-	}
-	closedir(dp);
-
-	qsort(filenames, n_files, sizeof(char*), (int (*)(const void *, const void *)) &strcmp);
-
-	for (size_t i = 0; i < n_files; i++) {
+	for (char **path = globbuf.gl_pathv; path != NULL; path++) {
 		pid_t pid = fork();
```
NeoRaider: Missing error check for fork.

admin/autoupdater/src/util.c
```diff
@@ -0,0 +1,97 @@
...
+
+
+void run_dir(const char *dir) {
+	char pat[strlen(dir) + 3];
+	sprintf(pat, "%s/*", dir);
+	glob_t globbuf;
+	if (glob(pat, 0, NULL, &globbuf))
+		return;
+
+	for (char **path = globbuf.gl_pathv; path != NULL; path++) {
```
NeoRaider: `for (size_t i = 0; i < globbuf.gl_pathc; i++)` (path is never NULL on musl-libc)

admin/autoupdater/src/util.c
```diff
@@ -0,0 +1,97 @@
...
+			fputs("autoupdater: warning: failed to fork: ", stderr);
+			perror(NULL);
+		} if (pid == 0) {
+			int null_fd = open("/dev/null", O_RDWR);
+			dup2(null_fd, 0);
+			dup2(null_fd, 1);
+			dup2(null_fd, 2);
+			close(null_fd);
+			execl(*path, *path, (char *)NULL);
+			fprintf(stderr, "autoupdater: warning: failed executing %s: ", *path);
```
NeoRaider: This error message is never printed, as stderr is closed before...

admin/autoupdater/src/util.c
NeoRaider: It would also be nice to add a call to access() here to check the file is executable before doing the fork (otherwise it will also try to exec the README files...)

admin/autoupdater/src/uclient.h
NeoRaider: It would be nice to put `*(void **)cl->priv` into parentheses.

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,144 @@
...
+		if (ret < 0) {
+			request_done(cl, UCLIENT_ERROR_REDIRECT_FAILED);
+			return;
+		}
+		if (ret > 0) {
+			uc_data(cl)->retries++;
+			return;
+		}
+	}
+	uc_data(cl)->retries = 0;
```
NeoRaider: In what situation would setting retries to 0 here ever have an effect?

admin/autoupdater/src/uclient.c
```diff
@@ -0,0 +1,144 @@
...
+		request_done(cl, UCLIENT_ERROR_TOO_MANY_REDIRECTS);
+		break;
+	default:
+		request_done(cl, UCLIENT_ERROR_STATUS_CODE | cl->status_code);
+	}
+}
+
+
+static void eof_cb(struct uclient *cl) {
+	request_done(cl, cl->data_eof? 0 : UCLIENT_ERROR_CONNECTION_RESET_PREMATURELY);
```
NeoRaider: Please add a space before the '?'.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,458 @@
...
+		OPTION_SIMULATE = 'n',
+		OPTION_FALLBACK = 256,
+	};
+
+	const struct option options[] = {
+		{"branch",     required_argument, NULL, OPTION_BRANCH},
+		{"force",      no_argument,       NULL, OPTION_FORCE},
+		{"fallback",   no_argument,       NULL, OPTION_FALLBACK},
+		{"no-upgrade", no_argument,       NULL, OPTION_SIMULATE},
+		{"simulate",   no_argument,       NULL, OPTION_SIMULATE},
```
jplitza: Maybe. My reasoning: I *hate* when the short option doesn't match the long one. And because `-n` is the canonical short option for "don't really do anything", I had to find something with n. But because `--no-upgrade` didn't sound right (because what does the autoupdater do if not update/upgrade?), I also introduced `--simulate`.

So ideally, someone would propose a `--n*` option that makes more sense than `--no-upgrade` and we drop `--simulate`.

admin/autoupdater/src/autoupdater.c
```diff
@@ -0,0 +1,458 @@
...
+	/* Close stdin and stderr to not give sysupgrade unneccessary SIGHUPs */
+	{
+		int null_fd = open("/dev/null", O_RDWR);
+		dup2(null_fd, 0);
+		dup2(null_fd, 2);
+		close(null_fd);
+	}
+
+	/* Unset FD_CLOEXEC so the lockfile stays locked during sysupgrade */
+	fcntl(lock_fd, F_SETFD, 0);
```
jplitza: But does it harm anybody? Maybe people want to backport the autoupdater to earlier OpenWrts or call an entirely different program for whatever reason. It feels cleaner to let the lockfile stay open across the exec.

admin/autoupdater/src/autoupdater.c
Sunz3r: how about with --dry-run like rsync?

admin/autoupdater/src/uclient.c
jplitza: Since we always use the `get_url()` function, which creates a new uclient instance for every call, it never has. Otherwise the redirects would accumulate over multiple URLs in a single instance, which uclient supports IIRC. And basically, I more or less borrowed this code from ufetch, and didn't think it was useless at the time. I'll remove it.

jplitza:
I think I addressed most of the issues adequately, but I wasn't able to test them yet.

admin/autoupdater/src/manifest.c
```diff
@@ -32,7 +32,8 @@
 #include <string.h>
 
 
-void free_manifest(struct manifest *m) {
+// only frees the data inside the manifest struct, not the struct itself!
+void free_manifest_data(struct manifest *m) {
```
NeoRaider: You didn't adjust the function prototype name.

admin/autoupdater/src/autoupdater.c
```diff
@@ -350,6 +350,8 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 		}
 	}
 
+	free_manifest_data(m);
```
NeoRaider: More important than freeing the manifest in the successful case would be to free it when false is returned, as that is the case where more mirrors are tried next.

admin/autoupdater/src/util.c
```diff
@@ -55,13 +58,20 @@ void run_dir(const char *dir) {
...
+			int wstatus;
+			if (waitpid(pid, &wstatus, 0) != pid) {
+				fprintf(stderr, "autoupdater: warning: failed waiting for child %d corresponding to %s: ", pid, globbuf.gl_pathv[i]);
+				perror(NULL);
+			}
+			else if (!WIFEXITED(wstatus)) {
+				fprintf(stderr, "autoupdater: warning: execution of %s exited abnormally\n", globbuf.gl_pathv[i]);
+			}
+			else if (WEXITSTATUS(wstatus)) {
+				fprintf(stderr, "autoupdater: warning: execution of %s exited with status code %d\n", WEXITSTATUS(wstatus));
```
NeoRaider: Missing `globbuf.gl_pathv[i]` argument.

I guess it would make sense to assign the path to a variable at the top of the loop body instead of writing `globbuf.gl_pathv[i]` each time.

admin/autoupdater/src/hexutil.c
```diff
@@ -51,8 +47,3 @@ int parsehex(void *buffer, const char *string, size_t len) {
 
 	return 1;
 }
-
-void hexdump(FILE *stream, void *buffer, size_t len) {
```
NeoRaider: The prototype must be removed as well.

admin/autoupdater/src/autoupdater.c
rotanid: --no-action ?

jplitza: I like `--no-action`! If nobody objects or comes up with a better idea, I'll implement that on the next occasion and remove the current options for simulation.

PR #169, opened 2017-09-13
libbabelhelper: add initial version

christf:
This will add a helper library to parse data from babel.
This is used by mmfd and the babel respondd module.


christf:
the commit hash in this make file is subject to change.

PR #171, opened 2017-09-26
add diagnostics about failed dlopen

t-8ch:


NeoRaider:
Thanks, the patch looks good, but the commit message should be something like:

    respondd: add diagnostics about failed dlsym

t-8ch:
@NeoRaider I adapted the code to handle errors for both dlopen and dlsym and fixed fixed the commit message.
Could you take another look at it?

net/respondd/src/respondd.c
```diff
@@ -183,11 +183,20 @@ static const struct respondd_provider_info * get_providers(const char *filename)
...
+
+	// clean a potential previous error
+	dlerror();
 
 	const struct respondd_provider_info *ret = dlsym(handle, "respondd_providers");
 	if (!ret) {
+		char *error = dlerror();
+		if (error) {
+			syslog(LOG_WARNING, "unable to load providers from '%s', ignoring: %s", path, error);
+		}
```
NeoRaider: I think the message should be logged whenever dlsym() returns NULL, not only when dlerror() returns something. So either we could just pass `dlerror() ?: "whatever error message we should use here..."`, or, even simpler, don't use dlerror() at all for the second message, as it wouldn't give a much more specific error message than "unable to load providers from '%s'" already does.

Also, use *filename* instead of *path* for the error messages.

PR #172, opened 2017-10-08
mmfd: bump version to pull in latest adaptions to libbabelhelper

christf:


christf:
as agreed, moving along with mmfd version bumps.

PR #178, opened 2017-12-18
tunneldigger: Update to newest upstream commit:

lcb01a:
- Sync package version numbers (0.4.0 = vNext)
- Added cmake related patches to Makefile

The new upstream version contains a lot of bug fixes and the latest commits use the cmake system.

### DIFF
--- a/net/tunneldigger/Makefile
+++ b/net/tunneldigger/Makefile
@@ -1,9 +1,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tunneldigger
-PKG_VERSION:=0.4.7
+PKG_VERSION:=0.4.0
 PKG_RELEASE:=1
-PKG_REV:=7cc92020cbafe5be2b24eb6bc943a65f151c2a18
+PKG_REV:=d7db350011076d6a83855d412885a29a7d142b6e 
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks; I was about to send a similar PR: :)

However, could you please update to d7db350011076d6a83855d412885a29a7d142b6e ?  This contains some more bugfixes for the client.

Quoting the upstream changelog (removing changes only affecting the server side):

*  Fixed compatibility with new Linux kernels on the broker side: New kernels force the l2tpv3 session ID to be unique system-wide, while old tunneldigger clients have a hard-coded ID of 1 for both ends of the tunnel. When a new client talks to a new broker, they will instead use a unique ID. Moreover, when the broker detects that it runs on an affected kernel, it reports maximal usage to old clients -- so if the client does usage-based selection, it will pick another broker.
* Added cmake buildsystem to the client.
* Improve client behavior on broker failure, and decrease some reconnect timeouts.

The first one is the most important here.  All nodes in a community have to be updated to a version with this fix, before newer kernels can be rolled out on the gateways.

lcb01a:
Updated the pull request :)

NeoRaider:
See also 57c67964e2c0e7604464f8b77e5ffbba2f8ab9ba.

I'll merge the v2017.1.x update after this got more feedback, as I don't have a tunneldigger setup for testing myself.

lcb01a:
I've build a version and tested it on a node. All is looking fine

PR #183, opened 2018-02-18
autoupdater: Fix nullpointer dereference on invalid URL

TobleMiner:
Previously supplying an invalid url for download would result in
uclient_new returning NULL and crash the autoupdater as soon as
cl->priv is accessed.

Signed-off-by: Tobias Schramm <tobleminer@gmail.com>

### DIFF
--- a/admin/autoupdater/src/uclient.c
+++ b/admin/autoupdater/src/uclient.c
@@ -161,6 +161,9 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
 	};
 
 	struct uclient *cl = uclient_new(url, NULL, &cb);
+	if (!cl)
+		goto err;
+
 	cl->priv = &d;
 	if (uclient_set_timeout(cl, TIMEOUT_MSEC))
 		goto err;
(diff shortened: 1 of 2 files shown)


admin/autoupdater/src/uclient.c
```diff
@@ -182,7 +185,8 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
 
 	return d.err_code;
 
-err:
+err_uclient:
 	uclient_free(cl);
```
NeoRaider: Let's just do
```C
if (cl)
  uclient_free(cl)
```
rather than introducing a new label.

admin/autoupdater/src/uclient.c
```diff
@@ -161,19 +161,22 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
 	};
 
 	struct uclient *cl = uclient_new(url, NULL, &cb);
+	if(!cl) {
+		goto err;
+	}
```
NeoRaider: Coding style: add space after `if`, remove `{}`.

TobleMiner:
@NeoRaider Done. Should be ok now.

PR #184, opened 2018-02-18
autoupdater: Fix various possible nullpointer dereferences

TobleMiner:
The following commits fix some possible nullpointer dereferences due to failing dynamic memory allocations.

I've split the fixes into quite a few separate commits to make reviewing easier; just let me know in case you prefer fewer commits (1 per file or somehing)

Signed-off-by: Tobias Schramm tobleminer@gmail.com

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -144,7 +144,8 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 
 	if (optind < argc) {
 		settings->n_mirrors = argc - optind;
-		settings->mirrors = malloc(settings->n_mirrors * sizeof(char *));
+		settings->mirrors = safe_malloc(settings->n_mirrors * sizeof(char *), "failed to allocate memory for mirror list");
+
 		for (int i = optind; i < argc; i++) {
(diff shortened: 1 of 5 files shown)


NeoRaider:
This adds a lot of boilerplate, and even tries to recover after an allocation failed in some places, which is not very useful in my opinion.

Let's add a few small helper functions like `safe_malloc()`, which do malloc, check the result, print an error message and `abort()` on failure; this way, the actual use sites aren't cluttered with error checks.

TobleMiner:
Having a safe malloc function sounds like a good idea.

Recovering from failing signature allocations while parsing the manifest probably didn't make a lot of sense in the first place. If the manifest would still have enough valid signatures after recovering chances are wo don't have enough free memory to store the upgrade anyways. Thus aborting would be perferctly fine there, too.

TobleMiner:
@NeoRaider I've rewritten the patches and updated the PR

NeoRaider:
Thanks, I've merged this with a few changes (in particular, I removed the error message argument to keep the calls simple and added a *safe_realloc* helper.

Also, util.h was missing a stddef.h include for *size_t*.

TobleMiner:
Uh, thanks for the fixups

PR #185, opened 2018-05-12
roamguide: add package

sargon:
Latest state of roamguide. A tool to solve the situation where the client thinks he "sees" the AP nicely, but the AP have problems listing to the client. In the End it increases the user experience.

We have tested this in several high density settings, and the user experience increased dramatically.

### DIFF
--- a/net/roamguide/files/etc/config/roamguide
+++ b/net/roamguide/files/etc/config/roamguide
@@ -1,0 +1,11 @@
+# repeat the following for every wireless AP device you like to guide.
+config roamguide
+   option device "client0" # wireless AP device we like to guide
+   list signal '-60' # first penality level
+   list signal '-77' # second penality level
+   list signal '-85' # final penality level
+   option bantime_base '500'    # on the first level ban for 500ms
+   option bantime_factor '2375' # on the second one for 2775ms,
+                                # and in the final 10000ms
+   option forget_time '600' # Forget about a client after 10s
+   option enabled '1' # set to '0' to disable this profile
--- a/net/roamguide/files/usr/lib/micron.d/roamguide
+++ b/net/roamguide/files/usr/lib/micron.d/roamguide
@@ -1,0 +1,1 @@
+* * * * * /usr/bin/roamguide
--- a/net/roamguide/Makefile
+++ b/net/roamguide/Makefile
@@ -1,0 +1,33 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=roamguide
+PKG_VERSION:=0.0.4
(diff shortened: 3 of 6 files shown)


net/roamguide/files/etc/config/roamguide
```diff
@@ -0,0 +1,11 @@
+# repeat the following for every wireless AP device you like to guide.
```
rotanid: i would prefer a solution where this package detects itself if there is a single-band or a dual-band device, which in the latter case would need two sections for client0 and client1

rotanid:
@sargon usually we ask for improvements before the PR is merged, but maybe you can put my mind at rest by answering my doubts:
how is this intended to be deployed? if i understand it correctly, it would only do its work on 5GHz on most dual band devices by default, as there's only a default config for client0, not for client1. this means, it can't fully work in an RF environment with mixed single-band/dual-band devices as on the dual-band devices it won't touch the clients connected to the 2,4GHz wifi VIF

net/roamguide/files/etc/config/roamguide
rubo77: I opened an issue here: https://github.com/sargon/gluon-sargon/issues/14

net/roamguide/files/usr/lib/micron.d/roamguide
```diff
@@ -0,0 +1 @@
+* * * * * /usr/bin/roamguide
```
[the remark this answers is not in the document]
rubo77: https://github.com/sargon/gluon-sargon/pull/15

rubo77:
TODO:

- Lua statt bash
- Unterstützung für 5GHz

net/roamguide/files/etc/config/roamguide
maurerle: This issue is said to be fixed in https://github.com/sargon/gluon-sargon/commit/c9539a04b171d246d9643608c59f1a91ddb2d911 (though it just creates both and fails silently on single band devices).

maurerle:
As this package is unlikely to be merged in the current state, but seemst to actually be used by at least ffki, you might like to add this to the [community-packages repository](https://github.com/freifunk-gluon/community-packages/) instead if this is still needed and used.

That way gives a better overview of packages used in communities, while having a lower barrier set @sargon.

rotanid:
as there was not comment by the author in over 6 years, we close this.
there is a similar approach in gluons community-packages repository called ffda-gluon-usteer

rubo77:
https://github.com/freifunk-gluon/community-packages/pull/19

PR #187, opened 2018-05-21
autoupdater: allow skipping the versioncheck

lemoer:
This commit introduces a new cli flag "--no-versioncheck"

TODO:
- [x] Compile check
- [x] Functional check

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -101,7 +101,7 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 		OPTION_HELP = 'h',
 		OPTION_NO_ACTION = 'n',
 		OPTION_FALLBACK = 256,
-		OPTION_NO_VERSIONCHECK = 255,
+		OPTION_NO_VERSIONCHECK = 257,
 	};
 
 	const struct option options[] = {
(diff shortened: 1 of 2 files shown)


admin/autoupdater/src/autoupdater.c
```diff
@@ -86,6 +86,7 @@ static void usage(void) {
 		"                       really flash a new firmware if one is available.\n\n"
 		"  --fallback           Upgrade if and only if the upgrade timespan of the new\n"
 		"                       version has passed for at least 24 hours.\n\n"
+		"  --no-versioncheck    Skip version check and allow downgrades therefore.\n\n"
```
christf: word order is uncommon. How about: "Skip version check to allow downgrades."?

christf:
I like the idea. Until now it was convention that upgrade scripts will not have to support a downgrade.
Should the user be warned about this? Maybe use autoupdater with -n, then sysupgrade manually, again with -n, then reconfigure the device or just tell the user that the updater will go ahead with the update but there is a chance it will explode into the users face? I agree that downgrading easily is a feature worth having and I am not sure about either of those options.

rotanid:
@christf i somehow doubt that @lemoer s intention is to really support downgrades.
i guess it's just about issues with the version comparison, not actual downgrades.
... maybe he corrects me or he will just leave a thumbs up ;-)

admin/autoupdater/src/autoupdater.c
NeoRaider: I'd prefer something like `--force-version` or `--force-downgrade`.

lemoer:
Test is still outstanding.

lemoer:
Test worked. I think, this can be merged now.

PR #191, opened 2018-07-11
ebtables-tiny: add ebt-nat and ebt-redirect for according targets

T-X:
The redirect and dnat target are needed for gluon-alt-esc-client to forward frames to the selected, alternative gateways.

### DIFF
--- a/net/ebtables-tiny/src/Makefile
+++ b/net/ebtables-tiny/src/Makefile
@@ -6,7 +6,7 @@ CFLAGS += -Iinclude
 CFLAGS += -Wall
 
 EXT_TABLES:=filter nat broute
-EXT_FUNC:=arp ip6 ip limit mark mark_m standard
+EXT_FUNC:=arp ip6 ip limit mark mark_m nat standard
 EXT_OBJS:=$(foreach T,$(EXT_FUNC),extensions/ebt_$(T).o) $(foreach T,$(EXT_TABLES),extensions/ebtable_$(T).o)
 
 OBJECTS2:=getethertype.o communication.o libebtc.o useful_functions.o ebtables.o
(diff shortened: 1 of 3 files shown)


christf:
@T-X Thank you for your work. 
Why do we need these modules?

christf:
The question was answered on IRC. Only thing missing is a decision on the merge of gluon-alt-esc.

mweinelt:
1. We are interested in how the package size (ipkg file) on MIPS increases with these additions.
2. Can you enable LTO for this package? See e.g. https://lists.infradead.org/pipermail/openwrt-devel/2019-November/019749.html

maurerle:
This is used by the parker-flavour from FFBS:
https://gitli.stratum0.org/ffbs/packages/-/commits/v2023.2.x-ffbs-next

Gluon-Parker is a friendly fork, which integrates L3 wireguard as mesh-vpn, while doing BATMAN only in local mesh clouds:
https://github.com/ffbs/gluon-parker

@SmithChart would it work for you if this gets merged, or do you require the other patches as well?
In this case a new PR probably makes more sense?

SmithChart:
> @SmithChart would it work for you if this gets merged, or do you require the other patches as well?
In this case a new PR probably makes more sense?

We currently have #275 open, but that is unrelated to this change. And it is not needed for parker to work. It was just a debugging aid.

In parker we only have the `ebt-nat` module: https://gitli.stratum0.org/ffbs/packages/-/commit/917ee1a3158d8bfaefc1bf85f1dfdbc759c3bec9

What is your reasoning to add `ebt-redirect` as well?

rotanid:
today we decided to close this, as it has no use currently for gluon, no update in years and will be obsolete due to the upcoming switch to nftables anyway.
@T-X agreed to that.

PR #192, opened 2018-07-31
uc: add new package

christf:
This adds a smaller alternative for a socat-use case: write a line of text to a unix socket and display the response.

This is used for prefixd, used as debugging tool for l3roamd, can be used as diagnostic for ddhcpd and fastd replacing their respective ctl commands. 

This PR moves a package from the babel software out of the main tree into the packages repository.


christf:
Readme is now added. The code of UC already states that it is free software that should suffice for the purpose of this PR. I agree that ithe license should be clarified.

PR #207, opened 2018-12-11
autoupdater: add check for manifest info

feckert:
With the option -c (--check) we only download the manifest and verify
them. If manifest is valid and a new firmware is available then result
is printed to stdout.

VERSION=<image_version>
SIZE=<image_size>

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -84,6 +84,8 @@ static void usage(void) {
 		"  -h, --help           Show this help.\n\n"
 		"  -n, --no-action      Download and validate the manifest as usual, but do not\n"
 		"                       really flash a new firmware if one is available.\n\n"
+		"  -c, --check          Validate the manifest as usual, but do not\n"
+		"                       download a new firmware.\n\n"
 		"  --fallback           Upgrade if and only if the upgrade timespan of the new\n"
 		"                       version has passed for at least 24 hours.\n\n"
 		"  --force-version      Skip version check to allow downgrades.\n\n"
@@ -100,6 +102,7 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 		OPTION_FORCE = 'f',
 		OPTION_HELP = 'h',
 		OPTION_NO_ACTION = 'n',
+		OPTION_CHECK = 'c',
 		OPTION_FALLBACK = 256,
 		OPTION_FORCE_VERSION = 257,
 	};
@@ -109,12 +112,13 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 		{"force",     no_argument,       NULL, OPTION_FORCE},
 		{"fallback",  no_argument,       NULL, OPTION_FALLBACK},
 		{"no-action", no_argument,       NULL, OPTION_NO_ACTION},
+		{"check",      no_argument,      NULL, OPTION_CHECK},
 		{"force-version", no_argument, NULL, OPTION_FORCE_VERSION},
 		{"help",      no_argument,       NULL, OPTION_HELP},
 	};
 
 	while (true) {
-		int c = getopt_long(argc, argv, "b:fhn", options, NULL);
+		int c = getopt_long(argc, argv, "b:fhnc", options, NULL);
 		if (c < 0)
 			break;
 
@@ -137,6 +141,10 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 
 		case OPTION_NO_ACTION:
 			settings->no_action = true;
+			break;
+
+		case OPTION_CHECK:
+			settings->check = true;
 			break;
 
 		case OPTION_FORCE_VERSION:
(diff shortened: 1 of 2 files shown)


admin/autoupdater/src/autoupdater.c
```diff
@@ -334,6 +344,18 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 		goto out;
 	}
 
+	if (s->check) {
+		dump_manifest(info_path, m);
+		printf(
+			"autoupdater: info: Aborting only manifest check requested.\n"
```
rotanid: maybe a comma after "Aborting"? sounds strange the way it is at the moment

admin/autoupdater/src/autoupdater.c
```diff
@@ -56,8 +56,10 @@
 static const char *const download_d_dir = "/usr/lib/autoupdater/download.d";
 static const char *const abort_d_dir = "/usr/lib/autoupdater/abort.d";
 static const char *const upgrade_d_dir = "/usr/lib/autoupdater/upgrade.d";
+static const char *const check_d_dir = "/usr/lib/autoupdater/check.d";
 static const char *const lockfile = "/var/lock/autoupdater.lock";
 static const char *const firmware_path = "/tmp/firmware.bin";
+static const char *const info_path = "/tmp/info.txt";
```
rotanid: i think a more verbose name would be good, like something that includes the package name and something about the content, maybe something like "autoupdater-fw-update-info.txt" ?

admin/autoupdater/src/manifest.c
```diff
@@ -47,6 +49,24 @@ void clear_manifest(struct manifest *m) {
 }
 
 
+void dump_manifest(const char *info_path, struct manifest *m) {
+	int fd = open(info_path, O_WRONLY|O_CREAT, 0600);
```
NeoRaider: * How about writing the data to stdout instead?
* Using fopen/fprintf (or printf in the case of stdout) would simlify this code a a lot (and is generally more efficient than using open/write, as it doesn't require a syscall for each function call)

admin/autoupdater/src/autoupdater.c
```diff
@@ -334,6 +343,17 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 		goto out;
 	}
 
+	if (s->check) {
+		printf(
+			"VERSION=%s\n"
+			"SIZE=%u\n",
```
NeoRaider: imagesize is defined as an ssize_t, so the correct conversion pattern would be `%zd`.

admin/autoupdater/src/autoupdater.c
```diff
@@ -334,6 +343,17 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 		goto out;
 	}
 
+	if (s->check) {
+		printf(
+			"VERSION=%s\n"
```
NeoRaider: Hmm, do we want to add single quotes around the version number, so the output becomes parsable as shell code even when weird characters appear in the version (excluding the single quote...)?

admin/autoupdater/src/autoupdater.c
```diff
@@ -286,7 +294,8 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 	sprintf(manifest_url, "%s/%s.manifest", mirror, s->branch);
 
 
-	printf("Retrieving manifest from %s ...\n", manifest_url);
+	if (!s->check)
+		printf("Retrieving manifest from %s ...\n", manifest_url);
```
NeoRaider: I think eventually we should move all human-readable output from stdout to stderr, but for now this is okay.

admin/autoupdater/src/autoupdater.c
```diff
@@ -109,12 +112,13 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 		{"force",     no_argument,       NULL, OPTION_FORCE},
 		{"fallback",  no_argument,       NULL, OPTION_FALLBACK},
 		{"no-action", no_argument,       NULL, OPTION_NO_ACTION},
+		{"check",      no_argument,      NULL, OPTION_CHECK},
```
NeoRaider: no_argument is not alligned correctly.

rotanid:
@feckert do you intend to fix the reported issues with this PR?

feckert:
Sorry i was busy in the last weeks but **yes** I will fix the requested changes hop fully in the next days

bobcanthelpyou:
Could someone explain the difference between `-n` and `-c`? Are these options equal? Could these be merged? Or `-n` be replaced by `-c`?

feckert:
on -c we download only the manifest and verify the sign.
on -n we download the manifest and download the firmware but do not flash

bobcanthelpyou:
Thank you for the clarification. The description of `-n` is somehow misleading as it doesn't say anything about the firmware download.

rubo77:
Where did you find a description of -n?

belzebub40k:
@rubo77 https://github.com/freifunk-gluon/packages/blob/master/admin/autoupdater/src/autoupdater.c#L85

feckert:
> @feckert do you intend to fix the reported issues with this PR?

The change doesn't make any sense from my side, because I don't use the tool anymore. I would therefore close the pull request.

rubo77:
Maybe you should just add info that -n  also downloads the firmware to the help text

rubo77:
https://github.com/freifunk-gluon/packages/pull/222

PR #209, opened 2018-12-18
autoupdater: fix version compare bugs

### LINKED ISSUE #208: autoupdater: Firmware version 1.7.0 not considered newer than 1.7.0~rc1
It used to be the case that (following Debian's handling of version numbers) firmware version 1.7.0 was considered newer than 1.7.0\~rc1.  However, with the latest gluon 2018.1.3, that does not seem to work any more: We have rolled out 1.7.0\~rc1 on our beta branch, and now we built 1.7.0, but even when doing a force update check the autoupdater just says "No new firmware available."

NeoRaider:
Fixes two bugs in the autoupdater (one regression, one long-standing issue).

Closes #208

### DIFF
--- a/admin/autoupdater/src/version.c
+++ b/admin/autoupdater/src/version.c
@@ -34,8 +34,10 @@ static int char_order(char c) {
 		return 0;
 	else if (isalpha(c))
 		return c;
-	else if (c == '~')
+	else if (c == '\0')
 		return -1;
+	else if (c == '~')
+		return -2;
 	else
 		return c + 256;
 }
--- a/admin/autoupdater/src/version.c
+++ b/admin/autoupdater/src/version.c
@@ -47,7 +47,7 @@ bool newer_than(const char *a, const char *b) {
 	if (b == NULL)
 		return true;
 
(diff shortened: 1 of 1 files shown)


admin/autoupdater/src/version.c
```diff
@@ -34,8 +34,10 @@ static int char_order(char c) {
 		return 0;
 	else if (isalpha(c))
 		return c;
-	else if (c == '~')
+	else if (c == 0)
```
jplitza: Everywhere else we use `'\0'` instead of `0` for comparison.
```suggestion
	else if (c == '\0')
```

NeoRaider:
@jplitza At first, I wanted to make a change similar to your fixup, but dpkg and opkg agree that trailing zeros should be ignored.

jplitza:
@NeoRaider Then disregard my commit, but make sure to document that behavior somewhere - e.g. in the commit message, that currently says otherwise 😉. But of course also in the actual documentation.

rubo77:
Is it also tested that 10>9?

for example update from 2018.2.0.9 to 2018.2.0.10

PR #210, opened 2019-01-09
autoupdater: allow skipping the signatures check

belzebub40k:
When deploying a new firmware a signer might want to test it on a variety of devices. At the moment this requires copying the image manually and flashing via sysupgrade or decreasing the required signatures in the autoupdater. This commit introduces a new cli flag "--force-signatures" which allows the user to skip the signature check.

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -86,6 +86,7 @@ static void usage(void) {
 		"                       really flash a new firmware if one is available.\n\n"
 		"  --fallback           Upgrade if and only if the upgrade timespan of the new\n"
 		"                       version has passed for at least 24 hours.\n\n"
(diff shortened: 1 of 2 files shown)


NeoRaider:
I'm generally not a fan of skipping signature checks; other tools I've worked with like [RAUC](https://rauc.readthedocs.io/en/latest/) don't allow it either.

In particular, for this to work, you would need to push your (not sufficiently signed) manifest to a branch that is already configured on devices.
- Either this branch is made for test images (in that case, configure your branch with a separate test firmware key and properly sign the manifest - that way you can also be safe against transmission errors)
- Or it is a branch that is normally not used for such test images (in this case, putting an unsigned manifest in it is a bad idea, as other devices using the branch will suddenly throw errors when trying to find new firmwares.

rubo77:
Also it is already possible anyway, by setting the needed signature count down via UCI

PR #213, opened 2019-02-18
batman-adv-legacy: remove batctl v2013.4

T-X:
With the netlink functions backported batman-adv-legacy works fine with
the upstream batctl. Therefore there is no more need to build and
install the old batctl v2013.4 version.


net/batman-adv-legacy/Makefile
```diff
@@ -112,8 +66,6 @@ endef
 
 define Build/Patch
```
NeoRaider: We should be able to remove all these explicit Build/* steps as well and use the defaults. Also drop the batctl patches in the *patches* subdir.

T-X:
Changelog v2:
* removed batctl patches
* remove explicit Build/{DoPatch,Patch,Prepare} instructions (still needed this Build/Compile variant, otherwise the build would fail)

NeoRaider:
>     * remove explicit Build/{DoPatch,Patch,Prepare} instructions (still needed this Build/Compile variant, otherwise the build would fail)
Just rename MAKE_BATMAN_ADV_ARGS to MAKE_FLAGS, then you can use the default Build/Compile defined in `include/package-defaults.mk`. You can also drop ARCH, which is already in MAKE_FLAGS by default.

T-X:
Changelog v3:
* renamed MAKE_BATMAN_ADV_ARGS to MAKE_FLAGS
* removed Build/Compile

NeoRaider:
> You can also drop ARCH, which is already in MAKE_FLAGS by default.

Oops, I just noticed that ARCH is defined in the defaults, but not with the correct value. Please add it back, then we can get this merged.

T-X:
Changelog v4:
* readded "$ARCH" in MAKE_FLAGS

PR #189, opened 2018-07-01
add respondd-module-lldp

genofire:
Could be usefull on wired networks (with a lot of switches) - to find the right neigbours.

PR #217, opened 2019-04-10
gluon independent respondd-wifi package

genofire:


rotanid:
@genofire have you changed anything besides the name?
@NeoRaider basically we diff if the code stays the same and merge it?

genofire:
Remove a lot of parts, which ware not necessary to run it in a minimal version to keep fit with the current state of gluon. (but yes, it is nearly the same then #188)

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+	struct iface_list *ifaces;
+	int wifi24 = 0, wifi5 = 0, clients_count;
+
+	result = json_object_new_object();
+	if (!result)
+		return NULL;
+
+	clients = json_object_new_object();
+	if (!clients) {
+		json_object_put(wireless);
```
NeoRaider: There is no *wireless* variable in this function, did you compile test this?

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+	if (!wireless) {
+		json_object_put(result);
+		return NULL;
+	}
+
+	ifaces = get_ifaces();
+
+	while (ifaces != NULL) {
+
+		if(ifaces->type != NL80211_IFTYPE_ADHOC)
```
NeoRaider: What about MESH_POINT?

Also, missing space before opening parenthesis.

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+			wifi24 += clients_count;
+		if (ifaces->frequency > 5000)
+			wifi5 += clients_count;
+next_statistics: ;
+		void *freeptr = ifaces;
+		ifaces = ifaces->next;
+		free(freeptr);
+	}
+
+	//TODO maybe skip: if (wifi24 >  0 || wifi5 > 0) {
```
NeoRaider: I think it would make sense to skip if no AP interfaces were found at all.

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+	json_object_object_add(clients, "wifi24", json_object_new_int(wifi24));
+	json_object_object_add(clients, "wifi5", json_object_new_int(wifi5));
+	json_object_object_add(result, "clients", clients);
```
NeoRaider: Hmm, I'm not sure if it really makes sense to count the clients in a generic respondd module after all. Since freifunk-gluon/gluon@b850fff7e4f4b938c512d81de31bfae872fa42e4, "wifi" is computed from "wifi24" and "wifi5" as well - which would need to be moved into this module as well; but the computation of "total" depends on "wifi" now.

A different approach would be to move this query into a reusable library included by both gluon-mesh-... respondd providers...

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+		if (!station)
+			goto next_neighbours;
+
+		if (!get_neighbours(station, ifaces->ifx)) {
+			json_object_put(station);
+			goto next_neighbours;
+		}
+
+		if (ifaces->frequency)
+			json_object_object_add(station, "frequency", json_object_new_int(ifaces->frequency));
```
NeoRaider: The frequency does not belong in the "neighbours" respondd object. If anything, nodeinfo would be the correct place (but let's not discuss extensions of the current format in this PR, but rather keep the exact output that is currently produced by gluon-mesh-*)

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+	if (!clients) {
+		json_object_put(wireless);
+		json_object_put(result);
+		return NULL;
+	}
+
+	ifaces = get_ifaces();
+	while (ifaces != NULL) {
+		if(ifaces->type != NL80211_IFTYPE_AP)
+			goto next_statistics;
```
NeoRaider: This would also include AP interfaces that are not used for mesh clients (for example the private-wifi interface). Another reason to only provide this as a library for the gluon-mesh-* respondd providers rather than an independent module.

net/respondd-wifi/src/respondd.c
```diff
@@ -0,0 +1,104 @@
...
+		clients_count = 0;
+		get_client_counts(&clients_count, ifaces->ifx);
+		if (ifaces->frequency < 5000)
+			wifi24 += clients_count;
+		if (ifaces->frequency > 5000)
+			wifi5 += clients_count;
+next_statistics: ;
+		void *freeptr = ifaces;
+		ifaces = ifaces->next;
+		free(freeptr);
```
NeoRaider: The memory handling in both loops processing the get_ifaces() output is quite ugly. I think a for_each_iface() taking a callback function would be much nicer, as we could avoid building a list of all interfaces in memory.

rotanid:
@genofire do you plan to work on the feedback you got?

rotanid:
> @genofire do you plan to work on the feedback you got?

any feedback possible? :)

genofire:
oh sry, miss it -> my fault - was just looking for #189 and thought it was not anymore possible to disable it in batman that it does not have any collision ....

I will work on it on Monday

PR #233, opened 2019-07-31
simple-tc: fix confusing program names in error messages

jluebbe:
Error messages used `uradvd` and usage still had the `gluon-` prefix.


mweinelt:
Ahh, the copy pasta is real!

Thanks!

PR #237, opened 2019-12-30
autoupdater: Fixed segfault on wrong long option

oszilloskop:
If an autoupdater long option (prefixed with '--') are unknown then a segmentation fault occurs.
Example:
```
root@node:~# autoupdater --xyz
Segmentation fault
```
Fix:
autoupdater option struct termination by using ```{0, 0, 0, 0}```

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -112,6 +112,7 @@ static void parse_args(int argc, char *argv[], struct settings *settings) {
 		{"no-action", no_argument,       NULL, OPTION_NO_ACTION},
 		{"force-version", no_argument, NULL, OPTION_FORCE_VERSION},
 		{"help",      no_argument,       NULL, OPTION_HELP},
+		{0, 0, 0, 0}
 	};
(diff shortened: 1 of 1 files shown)


NeoRaider:
I assume this will generate a compiler warning because 0 is implicitly casted to `const char *`. Just write `{}` instead of `{0,0,0,0}`.

oszilloskop:
After another rebuild I can't find a warning for autoupdater.c
 
I don't got deep inside the getop-functions, but the most of all manpages and the example from gnu.org shows me {0, 0, 0, 0}. Very rarely I have seen {NULL, 0, NULL, 0}.

The gnu.org example of using of getopt_long  -> https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Option-Example.html#Getopt-Long-Option-Example

blocktrron:
Both proposals should be valid for use with gcc. I personally prefer the solution from @NeoRaider, as it's independent from the underlying struct.

NeoRaider:
Thanks, merged with changes.

PR #240, opened 2021-01-01
Add new package wgpeerselector

lemoer:
wgpeerselector
==============

The wireguard peer selector is a daemon allowing to connect to only one
wireguard peer of a set of N peers. It does that by randomly picking one
peer, trying to connect and upon failure trying to connect to the next peer.
It is comparable to setting the "peer limit" to one in fastd. The daemon
is written in lua and hardwired to OpenWrt mechanisms.

Features:
- Status socket via UBUS.
- Routes for "Allowed IPs" are installed to the kernel.
- Synchronize time via NTP.
    - Currently necessary in wireguard to reestablish a connection after
      rebooting. [Reference](https://lists.zx2c4.com/pipermail/wireguard/2019-February/003850.html)
- Can change its unix group.

How does it work?
-----------------

The daemon is configured using its netifd proto `wgpeerselector`. The wireguard
interface is set up by the standard wireguard netifd proto `wireguard`.
Configuring peers via the `wireguard` proto is not necessary because this will
be done by wgpeerselector based on its uci config:

/etc/config/network:
```
config interface 'vpn'
	option proto 'wireguard'
	option fwmark '1'
	option private_key 'YOUR_PRIVATE_KEY_GOES_HERE'
	list addresses 'fe80::02a4:18ff:fe7e:a10d'
	option disabled '0'

config interface 'vpn_peerselector'
	option proto 'wgpeerselector'
	option transitive '1'
	option ifname 'vpn'
	option unix_group 'YOUR_UNIX_GROUP_GOES_HERE'   # this is optional
```

/etc/config/wgpeerselector:
```
config peer 'test_unknown_a'
	option enabled '1'
	option public_key 'mIDOdscl2R3Dq+YthxdTvvtH4D53VhawzEnmet+E7W0='
	list allowed_ips 'fe80::1/128'
	option ifname 'vpn'
	option endpoint 'hostname.example.com:51820'
```

Status information can be obtained via ubus:
```
root@platzhalter-525400123456:~# ubus call wgpeerselector.vpn status
{
	"peers": {
		"test_unknown_a": false,
		"test_unknown_b": false,
		"test_sn07": {
			"established": 1490
		}
	}
}
```

Algorithm:
----------

A connection is considered as "established" if the latest handshakes was
within the last 2.5 minutes.

Installing a peer:
1. Resolve its endpoint if necessary.
2. Randomly pick a remote IP.
3. Install the peer with that ip to the kernel.
4. Wait 5 seconds.
5. Check whether the connection is established.
    - If not, remove the peer from the kernel again.

Main Loop:
- Wait for the interface, if not existing.
- Try to synchronize time via NTP, if not done yet.
- If no connection is established:
	- Randomly pick a peer.
	- Try to establish a connection.
- If a connection is established:
	- Check every 5 seconds if the connection is still established.

Picking peers is done in such a way that all peers are tried once
before one peer is attempted for the second time. The same goes for the
DNS resolultion of the endpoints.

Warning: This algorithm only works if something is causing traffic into
the wireguard interface. This is due to the fact, that wireguard is only
doing handshakes, if it has to. While running mesh protocols on top of the
wireguard interface this is usually not a problem, since those protocols
periodically send Hello, OGM, ... packets.

### DIFF
--- a/net/wgpeerselector/Makefile
+++ b/net/wgpeerselector/Makefile
@@ -1,0 +1,25 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=wgpeerselector
+PKG_VERSION:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/wgpeerselector
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Wireguard peer selector daemon
+  DEPENDS:=+lua +wireguard-tools +libubox-lua +libubus-lua +luaposix
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/wgpeerselector/install
+	$(CP) ./files/* $(1)/
+endef
+
+$(eval $(call BuildPackage,wgpeerselector))
--- a/net/wgpeerselector/README.md
+++ b/net/wgpeerselector/README.md
@@ -1,0 +1,97 @@
+wgpeerselector
+==============
+
+The wireguard peer selector is a daemon allowing to connect to only one
+wireguard peer of a set of N peers. It does that by randomly picking one
+peer, trying to connect and on failure trying to connect to the next peer.
+It is comparable to setting the "peer limit" to one in fastd. The daemon
+is written in lua and hardwired to OpenWrt mechanisms.
+
+Features:
+- Status socket via UBUS.
+- Routes for "Allowed IPs" are installed to the kernel.
+- Synchronize time via NTP.
+    - Currently necessary in wireguard to reestablish a connection after
+      rebooting. [Reference](https://lists.zx2c4.com/pipermail/wireguard/2019-February/003850.html)
+- Can change its unix group.
+
+How does it work?
+-----------------
+
+The daemon is configured using its netifd proto `wgpeerselector`. The wireguard
+interface is set up by the standard wireguard netifd proto `wireguard`.
+Configuring peers via the `wireguard` proto is not necessary because this will
+be done by wgpeerselector based on its uci config:
+
+/etc/config/network:
+```
+config interface 'vpn'
+	option proto 'wireguard'
+	option fwmark '1'
+	option private_key 'YOUR_PRIVATE_KEY_GOES_HERE'
+	list addresses 'fe80::02a4:18ff:fe7e:a10d'
+	option disabled '0'
+
+config interface 'vpn_peerselector'
+	option proto 'wgpeerselector'
+	option transitive '1'
+	option ifname 'vpn'
+	option unix_group 'YOUR_UNIX_GROUP_GOES_HERE'   # this is optional
+```
+
+/etc/config/wgpeerselector:
+```
+config peer 'test_unknown_a'
+	option enabled '1'
+	option public_key 'mIDOdscl2R3Dq+YthxdTvvtH4D53VhawzEnmet+E7W0='
+	list allowed_ips 'fe80::1/128'
+	option ifname 'vpn'
+	option endpoint 'hostname.example.com:51820'
+```
+
+Status information can be obtained via ubus:
+```
+root@platzhalter-525400123456:~# ubus call wgpeerselector.vpn status
+{
+	"peers": {
+		"test_unknown_a": false,
+		"test_unknown_b": false,
+		"test_sn07": {
+			"established": 1490
+		}
+	}
+}
+```
+
+Algorithm:
+----------
+
+A connection is considered as "established" if the latest handshakes was
+within the last 2.5 minutes.
+
+Installing a peer:
+1. Resolve its endpoint if necessary.
+2. Randomly pick a remote IP.
+3. Install the peer with that ip to the kernel.
+4. Wait 5 seconds.
+5. Check whether the connection is established.
+    - If not, remove the peer from the kernel again.
+
+Main Loop:
+- Wait for the interface, if it does not exist.
+- Try to synchronize time via NTP, if not done yet.
+- If no connection is established:
+	- Randomly pick a peer.
+	- Try to establish a connection.
+- If a connection is established:
+	- Check every 5 seconds if the connection is still established.
+
+Picking peers is done in such a way that all peers are tried once
+before one peer is attempted for the second time. The same goes for the
+DNS resolultion of the endpoints.
+
+Warning: This algorithm only works if something is causing traffic into
+the wireguard interface. This is due to the fact, that wireguard is only
+doing handshakes, if it has to. While running mesh protocols on top of the
+wireguard interface this is usually not a problem, since those protocols
+periodically send Hello, OGM, ... packets.
--- a/net/wgpeerselector/files/usr/bin/wgpeerselector
+++ b/net/wgpeerselector/files/usr/bin/wgpeerselector
@@ -1,0 +1,501 @@
+#!/usr/bin/lua
+local unistd = require 'posix.unistd'
+local syslog = require 'posix.syslog'
+local grp = require 'posix.grp'
+local sys_sock = require 'posix.sys.socket'
+local sys_stat = require 'posix.sys.stat'
+local signal = require 'posix.signal'
+local time = require 'posix.time'
+local uloop = require 'uloop'
+local ubus = require 'ubus'
+local uci = require('simple-uci').cursor()
+
+local function log(level, message)
+	if _G.is_verbose then
+		local prefix = {
+			[syslog.LOG_ERR]  = "Error: ",
+			[syslog.LOG_CRIT] = "Critical: ",
+			[syslog.LOG_INFO] = "Info: ",
+			[syslog.LOG_WARNING] = "Warning: ",
+		}
+		io.stderr:write(prefix[level]..message .. '\n')
+	end
+
+	syslog.syslog(level, message)
+end
+
+local function execute(command)
+	local f = io.popen(command)
+	if not f then
+		return nil
+	end
+
+	local data = f:read('*a')
+	f:close()
+	return data
+end
+
+local function sleep(secs)
+	-- This function calls uloop.run() for 'secs' seconds, so we can handle i/o
+	-- during that time. But be aware, that an fd should be registered to uloop,
+	-- because otherwise this function uses 100% cpu. This might be a bug in
+	-- uloop.
+	local timer = uloop.timer(function() uloop.cancel() end)
+	timer:set(secs*1000)
+	uloop.run()
+end
+
+local WGPeer = {}
+
+function WGPeer:new(o)
+	o = o or {}   -- create object if user does not provide one
+	setmetatable(o, self)
+	self.__index = self
+	-- some defaults
+	o.rx_bytes = 0
+	o.tx_bytes = 0
+	o.latest_handshake = 0
+	o.established_at = 0
+	-- terminology:
+	---> endpoint (maybe) needs dns resolution
+	---> remote is already resolved
+	o.next_remotes = {}
+	return o
+end
+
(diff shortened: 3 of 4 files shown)


net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,505 @@
...
+	if os.execute(command) ~= 0 then
+		log(syslog.LOG_WARNING, "Couldn't remove peer " .. self.name .. " from " .. self.iface .. ".")
+	end
+	for _, allowed_ip in pairs(self.allowed_ips) do
+		if os.execute("ip route del '" .. allowed_ip .. "' dev '" .. self.iface .. "'") ~= 0 then
+			log(syslog.LOG_WARNING, "Couldn't del route " .. allowed_ip .. " from " .. self.iface .. ".")
+		end
+	end
+	self.established_at = 0
+end
```
blocktrron, line 145: `WGPeer:install_to_kernel()` returns boolean error codes, so should this function.

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,505 @@
...
+	self.__index = self
+	-- some defaults
+	self.queued_peers = {}
+	self.peers = {}
+	return o
+end
+
+function WGPeerSelector:load_peers_from_uci()
+	uci:foreach('wgpeerselector', 'peer', function(peer)
+		if tonumber(peer.enabled) == 1 and peer.ifname == self.iface and peer.public_key then
```
blocktrron, line 196: `uci:get_bool` might be a good idea, although it's two queries then.

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,505 @@
...
+	if #dns_results < 1 then
+		log(syslog.LOG_WARNING, "no (valid) dns result for ntp servers. skipping time sync.")
+		return false
+	end
+
+	local command = "ntpd -n -N -q -p '" .. table.concat(dns_results, "' -p '") .. "'"
+	local time_synced = os.execute(command) == 0
+
+	if not time_synced then
+		log(syslog.LOG_WARNING, "synchronizing time failed.")
```
blocktrron, line 305: Timesync failed

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,505 @@
...
+	else
+		log(syslog.LOG_INFO, "time is now synchronized.")
```
blocktrron, line 307: Timesync successful

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,505 @@
...
+		is_verbose = true
+	elseif arg[i] == "--help" or arg[i] == "-h" then
+		actions['help'] = true
+	end
+end
+
+if actions['version'] then
+	print("wgpeerselector v1")
+	os.exit(0)
+end
```
blocktrron, line 450: Ususally Gluon packages are not versioned, so I'd drop that.

lemoer: This doesn't really make sense to me. Versioning software is always a good idea, I would say. Furthermore it seems like they actually have. See e.g. [here](https://github.com/freifunk-gluon/packages/blob/master/net/respondd/Makefile#L4).

net/wgpeerselector/files/usr/bin/wgpeerselector
blocktrron: I don't see a change here, why is it marked as resolved?

net/wgpeerselector/files/usr/bin/wgpeerselector
blocktrron: Same as above

net/wgpeerselector/files/usr/bin/wgpeerselector
blocktrron: All Gluon packages are versioned as v1, as otherwise they are not built with theOpenWrt buildsystem.

So I would either version the package implicitly using the commit hash of the packages repo or drop this version command completely. Versioning of installed packages can be tracked using opkg with the first approach.

net/wgpeerselector/files/usr/bin/wgpeerselector
lemoer: Because I implemented things locally and thought, I'll push this in a few minutes, then things occurred...

net/wgpeerselector/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+PKG_NAME:=wgpeerselector
+PKG_VERSION:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/wgpeerselector
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Wireguard peer selector daemon
```
zx2c4, line 11: WireGuard has a capital G.

net/wgpeerselector/README.md
```diff
@@ -0,0 +1,97 @@
...
+peer, trying to connect and on failure trying to connect to the next peer.
+It is comparable to setting the "peer limit" to one in fastd. The daemon
+is written in lua and hardwired to OpenWrt mechanisms.
+
+Features:
+- Status socket via UBUS.
+- Routes for "Allowed IPs" are installed to the kernel.
+- Synchronize time via NTP.
+    - Currently necessary in wireguard to reestablish a connection after
+      rebooting. [Reference](https://lists.zx2c4.com/pipermail/wireguard/2019-February/003850.html)
```
zx2c4, line 15: The only _real_ requirement is that it's monotonic. So you could just store the time to flash at every N interval, restore from that at boot, and have at most N interval of downtime while the clock advances to wherever it was when power was cut.

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,501 @@
...
+
+local function sleep(secs)
+	-- This function calls uloop.run() for 'secs' seconds, so we can handle i/o
+	-- during that time. But be aware, that an fd should be registered to uloop,
+	-- because otherwise this function uses 100% cpu. This might be a bug in
+	-- uloop.
+	local timer = uloop.timer(function() uloop.cancel() end)
+	timer:set(secs*1000)
+	uloop.run()
```
zx2c4, line 45: Why not just use [`unistd.sleep`](https://luaposix.github.io/luaposix/modules/posix.unistd.html#sleep)? Anything special about uloop that makes it a better candidate?

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,501 @@
...
+
+function WGPeer:install_to_kernel()
+	local remote = self:get_next_remote()
+	if not remote then
+		return false
+	end
+
+	local command = "wg set '" .. self.iface .. "' peer '" .. self.public_key .. "' " ..
+	                "allowed-ips '" .. table.concat(self.allowed_ips, ",") .. "' " ..
+	                "endpoint '" .. remote .. "' " ..
```
zx2c4, line 115: `wg set ... endpoint ...` will resolve dns itself, so maybe you don't need get_next_remote? Unless you're really intent on doing some kind of round-robin multi homing, but usually that doesn't make sense for wireguard, since you rarely want to share private keys.

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,501 @@
...
+
+function WGPeerSelector:try_connect_to_peer(peer, timeout)
+	if not peer:install_to_kernel() then
+		return false
+	end
+
+	sleep(timeout)
+	peer:update_stats_from_kernel()
+
+	local connection_successful = peer:has_recent_handshake()
```
zx2c4, line 239: Checking for a recent handshake immediately after `wg set` implies that you have persistent keepalive enabled or something else generating traffic, right?

net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -0,0 +1,501 @@
...
+
+	while true do
+		if connected_peer then
+			state = 'established'
+		else
+			state = 'unconnected'
+		end
+
+		if not time_synchronized then
+			-- Wireguard requires time to be monotonic (always increasing). If
```
zx2c4, line 362: Capital G on WireGuard.

net/wgpeerselector/files/usr/bin/wgpeerselector
lemoer: Yes. That is exactly why we use it here. It allows us to serve the ubus api even when we are "asleep".

net/wgpeerselector/files/usr/bin/wgpeerselector
lemoer: This has mainly two reasons:
1. Better control of the error handling in case of dns failures.
2. To make connection attempts with IPv4 and IPv6 in alternation (IPv6, IPv4, IPv6, IPv4, IPv6, ...).

net/wgpeerselector/files/usr/bin/wgpeerselector
lemoer: Yes, for Gluon we have the mesh protocol that continuously generates traffic every X seconds. Also, I added a warning about this potential issue in the README ([here](https://github.com/freifunk-gluon/packages/pull/240/files/3a29b1fa0177ffcdecd3f15a3f51b448a71cb011#diff-e8d18fa5462091b44ec1c5a859daf30b8b1a745b6f4e5c742932cf22d529ca19R93-R97)).

net/wgpeerselector/README.md
lemoer: Of course it would also be possible to write the time to flash. Since neither Gluon nor OpenWrt currently do that, the only option for this package is to synchronize the time via ntp. I think, as we have attached the link to the "Reference" here, this is sufficient...

lemoer:
@zx2c4 Thank you for your review here. I hope I was able to address your concerns.

net/wgpeerselector/README.md
```diff
@@ -0,0 +1,97 @@
+wgpeerselector
+==============
+
+The wireguard peer selector is a daemon allowing to connect to only one
+wireguard peer of a set of N peers. It does that by randomingly picking one
```
blocktrron, line 5: randomingly --> randomly

net/wgpeerselector/README.md
```diff
@@ -0,0 +1,97 @@
...
+Installing a peer:
+1. Resolve its endpoint if necessary.
+2. Randomly pick a remote IP.
+3. Install the peer with that ip to the kernel.
+4. Wait 5 seconds.
+5. Check whether the connection is established.
+    - If not, remove the peer from the kernel again.
+
+Main Loop:
+- Wait for the inteface, if not existing.
```
blocktrron, line 81: inteface --> interface

blocktrron: if it does not exist.

lemoer:
@blocktrron Thanks for your suggestions. I implemented them and squashed everything.

maurerle:
@lemoer what is the license under which the wgpeerselector is published?
BSD-2-Clause? Could you clarify this please? :)

lemoer:
I would consider it MIT licenced.

maurerle:
@lemoer would you mind creating a PR which adds the PKG_LICENSE info line? I would prefer if this is done by the original author :)

PR #244, opened 2021-04-19
respondd-module-airtime: add phy

AiyionPrime:
This implements freifunk-gluon/gluon#2204.

The `wireless`-structure provided in statistics looks like this now:

```json
"wireless": [
    {
      "frequency": 5220,
      "active": 30461,
      "busy": 2057,
      "tx": 4,
      "idx": 1
    },
    {
      "frequency": 2412,
      "active": 31982,
      "busy": 7544,
      "tx": 156,
      "idx": 0
    }
  ],
```

There's a test router available at http://idx-test.n.ffh.zone .

*or as fallback* http://[2001:678:978:122:b64b:d6ff:fe22:4ebc]/cgi-bin/status

### DIFF
--- a/net/respondd-module-airtime/src/respondd.c
+++ b/net/respondd-module-airtime/src/respondd.c
@@ -7,7 +7,9 @@
 #include "ifaces.h"
 
 static struct json_object *respondd_provider_statistics(void) {
-	struct json_object *result, *wireless;
+	bool ok;
+	int newest_element_index;
+	struct json_object *last, *result, *wireless;
 	struct iface_list *ifaces;
 
 	result = json_object_new_object();
@@ -22,8 +24,12 @@ static struct json_object *respondd_provider_statistics(void) {
 
 	ifaces = get_ifaces();
 	while (ifaces != NULL) {
-		get_airtime(wireless, ifaces->ifx);
-
+		ok = get_airtime(wireless, ifaces->ifx);
+		if (ok) {
+			newest_element_index = json_object_array_length(wireless)-1;
+			last = json_object_array_get_idx(wireless, newest_element_index);
+			json_object_object_add(last, "phy", json_object_new_int(ifaces->wiphy));
+		}
 		void *freeptr = ifaces;
 		ifaces = ifaces->next;
 		free(freeptr);
--- a/net/respondd-module-airtime/README.md
+++ b/net/respondd-module-airtime/README.md
@@ -11,7 +11,8 @@ The format is the following:
         "busy": 46496566,
         "rx": 808415,
         "tx": 41711344,
-        "noise": 162
+        "noise": 162,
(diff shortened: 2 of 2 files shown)


AiyionPrime:
Thanks @blocktrron for `wiphy`.

net/respondd-module-airtime/src/respondd.c
```diff
@@ -22,8 +24,12 @@ static struct json_object *respondd_provider_statistics(void) {
 
 	ifaces = get_ifaces();
 	while (ifaces != NULL) {
-		get_airtime(wireless, ifaces->ifx);
-
+		ok = get_airtime(wireless, ifaces->ifx);
+		if (ok) {
+			newest_element_index = json_object_array_length(wireless)-1;
+			last = json_object_array_get_idx(wireless, newest_element_index);
+			json_object_object_add(last, "idx", json_object_new_int(ifaces->wiphy));
```
mweinelt, line 31:
```suggestion
			json_object_object_add(last, "wiphy", json_object_new_int(ifaces->wiphy));
```

Or maybe just `phy`? `idx` is just a bit too generic from my POV.

blocktrron: +1 for `phy`

AiyionPrime: Thats fine with me; resolved in d630d87799846a5bf2c6cc096f336360b5b660e2

AiyionPrime:
This is ready for review again.

blocktrron (approved):

net/respondd-module-airtime/src/respondd.c
```diff
@@ -22,8 +24,12 @@ static struct json_object *respondd_provider_statistics(void) {
 
 	ifaces = get_ifaces();
 	while (ifaces != NULL) {
-		get_airtime(wireless, ifaces->ifx);
-
+		ok = get_airtime(wireless, ifaces->ifx);
+		if (ok) {
+			newest_element_index = json_object_array_length(wireless)-1;
```
NeoRaider, line 29: Hmm, at a glance it seems that the return code of `get_airtime` is pretty useless - in some error paths it will return true even when no object has been added to the array. This can lead to `newest_element_index` pointing at a previous entry, or even worse, -1.

I think it would be better to change `get_airtime` to add the index instead of added the element afterwards.

AiyionPrime: I think fixing the return code of get_airtime should be handled in a new issue;

There were three possible places where one could have added the phy info.
The one you propose does just replicate the return code of the wrapped `nl_send_dump`.
At that place one would have to add the info the same way one does in respond.c, but would have to add the wiphy int as parameter for `get_airtime`in some way.

PR #246, opened 2021-06-21
autoupdater: Add Gluon release to HTTP header

citronalco:
Send the access point's currently installed Gluon version in the HTTP header "X-GLUON-VERSION".

Resolves https://github.com/freifunk-gluon/gluon/issues/2232

### DIFF
--- a/admin/autoupdater/src/uclient.c
+++ b/admin/autoupdater/src/uclient.c
@@ -151,7 +151,7 @@ ssize_t uclient_read_account(struct uclient *cl, char *buf, int len) {
 }
 
 
-int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data, ssize_t len) {
+int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data, ssize_t len, const char *firmware_version) {
 	struct uclient_data d = { .custom = cb_data, .length = len };
 	struct uclient_cb cb = {
 		.header_done = header_done_cb,
@@ -175,6 +175,10 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
 		goto err;
 	if (uclient_http_set_header(cl, "User-Agent", user_agent))
 		goto err;
+	if (firmware_version != NULL) {
+		if (uclient_http_set_header(cl, "X-Firmware-Version", firmware_version))
+			goto err;
+	}
 	if (uclient_request(cl))
 		goto err;
 	uloop_run();
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -292,7 +292,7 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 
 	/* Download manifest */
 	ecdsa_sha256_init(&m->hash_ctx);
-	int err_code = get_url(manifest_url, recv_manifest_cb, &manifest_ctx, -1);
+	int err_code = get_url(manifest_url, recv_manifest_cb, &manifest_ctx, -1, s->old_version);
 	if (err_code != 0) {
 		fprintf(stderr, "autoupdater: warning: error downloading manifest: %s\n", uclient_get_errmsg(err_code));
 		goto out;
@@ -358,7 +358,7 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 		char image_url[strlen(mirror) + strlen(m->image_filename) + 2];
 		sprintf(image_url, "%s/%s", mirror, m->image_filename);
 		ecdsa_sha256_init(&image_ctx.hash_ctx);
-		int err_code = get_url(image_url, &recv_image_cb, &image_ctx, m->imagesize);
+		int err_code = get_url(image_url, &recv_image_cb, &image_ctx, m->imagesize, s->old_version);
 		puts("");
 		if (err_code != 0) {
 			fprintf(stderr, "autoupdater: warning: error downloading image: %s\n", uclient_get_errmsg(err_code));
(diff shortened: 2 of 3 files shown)


NeoRaider:
Thanks for the PR, however the current implementation doesn't fit into the "packages" repo, which only contains packages that might be moved to another OpenWrt feed at some point, and thus may not contain any Gluon-specific features.

There is also one issue regarding the behaviour of the code: What the autoupdater considers an image "version" is actually the Gluon "release" and should use the "X-Gluon-Release" header.

One option to implement this without any Gluon-specific features in the autoupdater itself would be the following:
- Introduce a new UCI section for custom HTTP headers in `/etc/config/autoupdater`
- Add an upgrade script to the `gluon-autoupdater` package that creates this UCI section

I'm currently unsure whether such an implementation would actually be necessary - maybe what this PR does is already sufficient for most use cases, and we would just need to rename the header ("X-Firmware-Version"?).

citronalco:
- Changed the name of the header from "X-Gluon-Version" to "X-Gluon-Release"
- Updated this PR's title to reflect this change
- Autoupdater no longer tries to set the "X-Gluon-Release" header if the Gluon release could not get detected, so it does not depend on Gluon.

rubo77:
> s, and we would just need to rename the header ("X-Firmware-Version"?).
If Openwrt also stores the firmwares version the, this sounds the better solution

citronalco:
I think I got the point, finally.
Calling the header "X-Firmware-Version" really makes more sense. I changed the code accordingly.

NeoRaider:
> > we would just need to rename the header ("X-Firmware-Version"?).
> 
> If Openwrt also stores the firmwares version there, this sounds the better solution

This has absolutely nothing to do with OpenWrt versions.

NeoRaider:
The implementation looks good now.

@blocktrron: You were working on an implementation yourself, but I think this PR is sufficient. Thoughts?

rubo77:
> might be moved to another OpenWrt feed at some point, and thus may not contain any Gluon-specific features.

What are you referring to, of not Openwrt?

citronalco:
Before this gathers too much dust:
Is there anything left I could do to get this merged?

NeoRaider:
In my opinion this is fine and can be merged, but I'd like to have @blocktrron's opinion as well.

blocktrron:
Sorry, I've forgotten about this.

I've had this alternative approach gathering dust for quite some time. It has the benefit of allowing communities to add additional custom headers (site, domain) and keep the autoupdater decoupled from Gluon.

https://github.com/blocktrron/gluon/tree/pr-autoupdater-custom-headers

NeoRaider:
> Sorry, I've forgotten about this.
> 
> I've had this alternative approach gathering dust for quite some time. It has the benefit of allowing communities to add additional custom headers (site, domain) and keep the autoupdater decoupled from Gluon.
> 
> https://github.com/blocktrron/gluon/tree/pr-autoupdater-custom-headers

Hmm, in any case all additional headers that might be useful would need to be added by default - as soon as you need a firmware update to add HTTP headers for another firmware update, you can also just write different mirror URLs in the domain configs by updating the firmware. I'm unsure how useful additional headers are.

blocktrron:
Usefulness has to decide everyone based on it's use-case. I'd prefer the UCI option, as it does not make the autoupdater more gluon specific.

NeoRaider:
> Usefulness has to decide everyone based on it's use-case. I'd prefer the UCI option, as it does not make the autoupdater more gluon specific.

@blocktrron Please have a look at this PR again, it isn't Gluon-specific at all. It just uses the version number that is also used for everything else by the autoupdater.

blocktrron:
Sorry, I've mistaken this for a different approach we've talked about.

In this case, I do like the approach from this PR more, as it does not introduce additional boilerplate in the core repository.

T-X:
@citronalco Just one general comment. It's probably obvious to everyone involved here, but in general it would be good to have one or two sentences regarding the "why" in the commit message.

Out of curiousity, are there applications which are already using such information? Or is this just for logging purposes on the webserver?

citronalco:
@T-X  Basically this could be used to provide different firmware files, depending on the device's current firmware.
The web server could check the devices' firmware version in the header, and provide different firmware updates depending the devices's current Gluon version.

A use case is described in the linked Feature Request (https://github.com/freifunk-gluon/gluon/issues/2232):
Communities sign their firmware, probably with keys of some of their members. Over the years members come and go, and so do firmware signatures.
So when updating from very old firmware release it might not be possible to update to the most current release immediately: The device's firmware could only accept signatures that are no longer used to sign the most current firmware - one has to do 1:n update steps in between, to follow the changes in the required/provided signatures.

Another use case would be something like this:
The Gluon 2021.1 release notes state: "Upgrades to v2021.1 and later releases are only supported from releases v2018.2 and later". Currently there is now easy way to e.g. a device running Gluon 2017 to update to Gluon 2019 before updating it to 2020.1.
Surely this Pull Request comes to late for this, but it could help when this happens again in the future:
One could configure the firmware web server to check the version in the HTTP header, and provide different firmware releases depending on the devices currently used release.

T-X:
@citronalco Ah, awesome idea! Thanks for the clarification! (and you're right, I missed reading the linked ticket, thanks for clarifying anyways)

admin/autoupdater/src/uclient.c
```diff
@@ -175,6 +175,10 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
 		goto err;
 	if (uclient_http_set_header(cl, "User-Agent", user_agent))
 		goto err;
+	if (firmware_version != NULL) {
+	    if (uclient_http_set_header(cl, "X-Firmware-Version", firmware_version))
+		    goto err;
```
NeoRaider, line 180: The indentation is off here, there should be one tab per level and no spaces.

citronalco: ok, fixed.

PR #247, opened 2021-07-29
autoupdater: verify the image before upgrading

### LINKED ISSUE #193: autoupdater: doesn't care about sysupgrade exit code
this issue was detected during https://github.com/freifunk-gluon/gluon/issues/1496
currently, the new autoupdater doesn't handle the exit code of the sysupgrade script it calls.
therefore, it doesn't put the node back into an accessible or working state after sysupgrade fails - a manual power cycle is necessary in this case.

grische:
Let sysupgrade run a --test upgrade to verify that the image is
compatible before attempting an upgrade.
This fixes an issue where a router can get stuck without network
connectivity when a a remote autoupgrade was triggered.

Fixes #193

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -380,10 +380,12 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 
 	/* Test the image upgrade (issue #193) */
 	{
+		static const char *const exec_builtin = "exec ";
 		static const char *const test_option = " --test ";
 
-		char buf[strlen(sysupgrade_path) + strlen(test_option) + strlen(firmware_path) + 1];
-		strcpy(buf, sysupgrade_path);
+		char buf[strlen(exec_builtin) + strlen(sysupgrade_path) + strlen(test_option) + strlen(firmware_path) + 1];
+		strcpy(buf, exec_builtin);
+		strcat(buf, sysupgrade_path);
 		strcat(buf, test_option);
 		strcat(buf, firmware_path);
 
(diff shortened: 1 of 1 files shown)


blocktrron:
Okay, this is embarassing - I've created https://github.com/freifunk-gluon/packages/pull/248 and saw your PR right after.

Maybe @NeoRaider can evaluate which approach he likes more. The biggest difference as far as i can tell is the `system` call creates an additional shell process.

NeoRaider:
I think `system` is the nicer approach here, as it keeps the code simpler. The additional shell process can be avoided by adding `exec ` to the beginning of the command passed to `system`.

grische:
@NeoRaider any chance we can get this merged? This problem is present on some routers which try to "autoupgrade" from a OpenWRT version 21.02 to an older one, as some router names seem to have changed.

NeoRaider:
@grische Please add the `exec` as I suggested, the rest looks fine.

admin/autoupdater/src/autoupdater.c
```diff
@@ -378,6 +378,24 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
...
 	}
 
+	/* Test the image upgrade (issue #193) */
+	{
+		static const char *const test_option = " --test ";
+		static const char *const exec_builtin = "exec ";
+
+		char buf[strlen(exec_builtin) + strlen(sysupgrade_path) + strlen(test_option) + strlen(firmware_path) + 1];
+		strcpy(buf, exec_builtin);
+		strcpy(buf, sysupgrade_path);
```
NeoRaider, line 388:
```suggestion
		strcat(buf, sysupgrade_path);
```

grische: Fixed, thanks!

PR #248, opened 2021-08-18
autoupdater: execute sysupgrade test before network shutdown

blocktrron:
Currently the autoupdater stops network prior starting the sysupgrade
process.

This has the issue of leaving the device in a state with stopped network
in case the sysupgrade process fails, as execl does not spawn a
subprocess but instead replaces the current process with another one, in
our case the sysupgrade process.

This commit modifies the autoupdater behavior in a way it starts
sysupgrade in test mode prior stopping network. This way, common causes
for failing the sysupgrade process can be detected in a
remote-recoverable manner, most prominently a failed platform check /
incompatible images.

Signed-off-by: David Bauer <mail@david-bauer.net>

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -48,6 +48,7 @@
 #include <sys/types.h>
 #include <sys/file.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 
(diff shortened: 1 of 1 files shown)


NeoRaider:
Alternative PR #247 merged.

PR #250, opened 2021-09-30
wgpeerselector - keep datatype on ubus status

genofire:
an other datatype in same json structure makes problem an marshalling:
here an object and an boolean.

if there is no null posible, please add an empty object - this suggestes solution here.
or skip the else type

### DIFF
--- a/net/wgpeerselector/files/usr/bin/wgpeerselector
+++ b/net/wgpeerselector/files/usr/bin/wgpeerselector
@@ -323,9 +323,7 @@ function WGPeerSelector:status()
 				established = established
 			}
 		else
-			result.peers[peer.name] = false -- Fastd uses "null" here, but unfortunately
-			                                -- lua does not support setting table_keys to
-			                                -- nil.
+			result.peers[peer.name] = { }
 		end
 	end
 


NeoRaider:
Let's just leave out the field then, an empty object doesn't make any sense.

A more involved solution would be to add a special `null` userdata to the `lua-jsonc` library, but I don't think that's necessary.

NeoRaider:
There's still a typo in the commit title. Also, please change the title to use a colon for the package name prefix (`wgpeerselector: ...`).

AiyionPrime:
A bird just mentioned "c-json kann das" ;)

lemoer:
> Let's just leave out the field then, [...]

No, this is a bad idea. Then unestablished connections would no longer be displayed on the status page.

> [...] an empty object doesn't make any sense.

I agree. Let's find another solution.

AiyionPrime:
Does one of you know, whether "upstream"-jsonc supports table assignments with null?

https://github.com/openwrt/luci/blob/master/libs/luci-lib-jsonc/src/jsonc.c

NeoRaider:
It really doesn't matter which JSON library can or cannot do this, as this is never turned into JSON at all, it is directly marshaled to ubus. libubus-lua is missing such a feature as well.

I just looked at the surrounding code again. I think an empty object may be nice here after all.

genofire:
so is this commit okay so, or should i change something?

PR #254, opened 2022-01-20
pretty-hostname: instantly apply hostname

mkg20001:
This makes it work like the CLI

### DIFF
--- a/admin/pretty-hostname/files/pretty_hostname.lua
+++ b/admin/pretty-hostname/files/pretty_hostname.lua
@@ -36,6 +36,10 @@ function M.set(uci, pretty_hostname)
 		uci:set('system', system, 'pretty_hostname', pretty_hostname)
 	end
 
+	local kernel = assert(io.open('/proc/sys/kernel/hostname', 'w'))
+	kernel:write(hostname)
+	kernel:close()
+
 	uci:set('system', system, 'hostname', hostname)
 end
 


mweinelt (approved):

NeoRaider:
I don't really think this makes sense - the Lua library was intentionally more low-level than the CLI util; I'd generally expect UCI-based configuration not to have any effect until the next reboot / service restart. This is just my opinion though, so I won't ask for a revert, as it seems that @mweinelt is fine with the change.

mkg20001:
Apply on restart doesn't really exist for the hostname, as there isn't a service to restart afaik. Adding it prob would be overkill so this sounds the sanest solution to me.

NeoRaider:
The question is what the usecase is - at the moment all users of this library are part of the config mode or upgrade scripts, so the old hostname isn't really observable. Do you have custom packages that make use of the Lua library?

mkg20001:
Yes. I've written one for funkfeuer graz that pulls data from manman.graz.funkfeuer.at (their node DB that contains OLSR addresses). It updates the ips & hostname and runs reconfigure&reload afterwards during runtime. So having the hostname update be visible instantly is cool.

(Also I'm currently working on OLSR support. https://github.com/mkg20001/gluon which ig we'd have to talk about, since upstreaming that is going to be effort)

PR #256, opened 2022-03-22
respondd: add provider "respondd-providers" to list registrations

T-X:
Add an internal respondd provider "respondd-providers" which allows a
respondd querier to query which providers are registered on a node.
Example output:

```
{
	"respondd-providers": [
		"neighbours",
		"nodeinfo",
		"statistics",
		"statistics-extended"
	]
}
```

This can be useful for providers which are known to be bulky, where the
actual data is better queried via HTTP/TCP instead. But where such a
provider is only present on a few nodes, so you'd want to avoid
trying each node via HTTP.

Signed-off-by: Linus Lüssing \<linus.luessing@c0d3.blue\>

### DIFF
--- a/net/respondd/src/respondd.c
+++ b/net/respondd/src/respondd.c
@@ -78,6 +78,11 @@ struct provider_list {
 	respondd_provider provider;
 };
 
+struct key_list {
+	struct key_list *next;
+	char *key;
+};
+
 struct request_type {
 	struct provider_list *providers;
 
@@ -101,6 +106,7 @@ struct request_schedule {
 
 static int64_t now;
 static struct hsearch_data htab;
+static struct key_list *keys;
 
 
 static struct json_object * merge_json(struct json_object *a, struct json_object *b);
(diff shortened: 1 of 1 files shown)


net/respondd/src/respondd.c
```diff
@@ -272,6 +278,23 @@ static void load_cache_time(struct request_type *r, const char *name) {
 
 }
 
+static void add_key(const char *key) {
+	struct key_list **pos, *pentry;
+	int i = 0;
```
maurerle, line 283: this is unused

net/respondd/src/respondd.c
```diff
@@ -306,6 +329,31 @@ static void add_provider(const char *name, const struct respondd_provider_info *
...
+	return ret;
+}
+
+static void add_providers_provider(void) {
+	const struct respondd_provider_info providers[] = {
+		{"respondd-providers", respondd_provider_providers},
+		{}
+	};
+
+	add_provider("respondd", providers);
```
maurerle, line 354: Wouldn't it make sense to bundle the add_key and add_provider together?
And also call add_key for respondd-providers so that it includes itself in the list?
Or should it return all "other" providers?

net/respondd/src/respondd.c
```diff
@@ -339,10 +387,14 @@ static void load_providers(const char *path) {
...
 			continue;
 
-		for (; providers->request; providers++)
+		for (; providers->request; providers++) {
 			add_provider(ent->d_name, providers);
+			add_key(providers->request);
+		}
 	}
 
+	add_providers_provider();
```
maurerle, line 396: passing multiple -d directories registers duplicate function pointers for "respondd-providers" - so the `add_providers_provider` should be moved to main instead

maurerle (changes requested):
I left some feedback - without this, this looks good.

net/respondd/src/respondd.c
maurerle: maybe we should add a CI with `-Werror` to spot such things?

PR #257, opened 2022-05-05
change Github URLs from git:// to https://

mmalte:
Github dropped git:// support, batman-adv-legacy is the only affected repo in v2019.1.x

Needed to build images wirh CVE-2022-24884 fix.

PR #258, opened 2022-05-23
Fix PKG_SOURCE_URL of batman-adv-legacy and tunneldigger

bobidle:
The build of Gluon is failing cause of an issue with the PKG_SOURCE_URL of some packages.

It is not possible to clone the source repositories using the git:// prefix, e.g. local tests:
```
$ git clone git://github.com/freifunk-gluon/batman-adv-legacy.git
Cloning into 'batman-adv-legacy'...
fatal: unable to connect to github.com:
github.com[0: 140.82.121.3]: errno=Connection timed out
```

```
$ git clone git://github.com/wlanslovenija/tunneldigger.git
Cloning into 'tunneldigger'...
fatal: unable to connect to github.com:
github.com[0: 140.82.121.3]: errno=Connection timed out
```

Would be great to update the old package version of v2018.2.x to build the latest security fix of Gluon easily.

### DIFF
--- a/net/batman-adv-legacy/Makefile
+++ b/net/batman-adv-legacy/Makefile
@@ -17,7 +17,7 @@ BATCTL_VERSION:=2013.4.0
 BATCTL_MD5SUM:=42e269cc710bbc9a8fd17628201d4258
 
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=git://github.com/freifunk-gluon/batman-adv-legacy.git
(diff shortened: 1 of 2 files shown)


bobidle:
Related issue #257

NeoRaider:
As proposed in https://github.com/freifunk-gluon/packages/pull/257#issuecomment-1118860174, our recommended solution is to use `git config --global url."https://github.com/".insteadOf git://github.com/`, as we don't intend to put more than the absolute minimum effort into end-of-life Gluon versions.

PR #259, opened 2022-06-27
sse-multiplexd.c: fix leaking pipe FDs

herbetom:
sse-multiplexd leaks pipe FDs to its child processes, so newer children keep the older ones alive when they should have received a SIGPIPE already

fixes freifunk-gluon/gluon#2468

@NeoRaider I don't really know what to write as a commit message, you are very welcome to open a PR yourself, edit this one or whatever. My intention is just to get the fix upstream. In today's FFRN nightly I had already patched it in, but of course I would like to get rid of the downstream patch quickly.

### DIFF
--- a/net/sse-multiplex/src/sse-multiplexd.c
+++ b/net/sse-multiplex/src/sse-multiplexd.c
@@ -97,7 +97,7 @@ struct provider {
 
 static int run_command(const char *command) {
         int pipefd[2];
-        if (pipe(pipefd) < 0) {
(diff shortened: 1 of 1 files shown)


NeoRaider:
Thanks - I couldn't edit the commit message of your PR, so I had to open a new one: #260

herbetom:
Ah, okay, editing doesn't seem possible if the PR comes from an org account. Good to know. Either way, i'm happy if the issue get's resolved.

PR #260, opened 2022-07-02
sse-multiplex: fix pipe FD leaks

NeoRaider:
sse-multiplexd leaked pipe FDs to its child processes, so newer children
keep the older ones alive when they should have received a SIGPIPE already.

Closes freifunk-gluon/gluon#2468


rotanid:
@NeoRaider the "closes" is not 100% correct as we need to bump the modules in the main gluon repository after merging here.

i'll merge it now, as you did not mark it as draft and it's the same code @herbetom already published and consider it "reviewed"

PR #263, opened 2023-06-14
autoupdater: delete /etc/opkg/keys on autoupdater upgrade

maurerle:
- does trigger on autoupdate after checking that the image is correct and autoupgrade should be done

fixes https://github.com/freifunk-gluon/gluon/issues/2665


maurerle:
This has been tested forcing an autoupdate (the allow downgrade option) on my DAP-X1860, which successfully removed the old keys

maurerle:
@neocturne perfers to fix this in OpenWrt, by having a config symbol that controls whether `/etc/opkg/keys/` is added to `Package/opkg/conffiles`

PR #267, opened 2023-10-13
simple-tc: fix kmod dependencies

neocturne:
simple-tc hasn't been working since OpenWrt 22.03, as act_police was moved to a separate package.

While we're at it, we can also clean up the other dependencies:

- Remove the obsolete conditional dependencies, these were for GLUON_SPECIALIZE_KERNEL
- cls_basic was moved to kmod-sched-core, so we don't need kmod-sched anymore

Fixes https://github.com/freifunk-gluon/gluon/issues/3017

### DIFF
--- a/net/simple-tc/Makefile
+++ b/net/simple-tc/Makefile
@@ -1,9 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=simple-tc
-PKG_VERSION:=2
-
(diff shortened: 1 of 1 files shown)


maurerle (approved):

maurerle:
I can confirm that it works properly when `opkg update && opkg install kmod-sched-act-polic` has been executed.

rotanid:
we probably should apply this update to master and v2023.1.x

neocturne:
> we probably should apply this update to master and v2023.1.x

For now, both use the same branch of gluon/packages

PR #268, opened 2023-10-17
libplatforminfo: handle armsr target

Kistelini:
This is required for https://github.com/freifunk-gluon/gluon/pull/3027.


neocturne:
I think it would make sense to move `x86.c` to something like `template/subtarget.c` and symlink it for both `x86.c` and `armsr.c` instead of duplicating the file

neocturne:
Thanks!

PR #269, opened 2023-10-29
ebtables-tiny: fix lockfile function

jluebbe:
flock without LOCK_NB never returns on a conflicting lock, so this function doesn't conform to its comment.

Also close the lock file on error, so we can keep retrying.

### DIFF
--- a/net/ebtables-tiny/src/libebtc.c
+++ b/net/ebtables-tiny/src/libebtc.c
@@ -127,6 +127,7 @@ void ebt_list_extensions()
 static int lock_file()
 {
 	int fd, try = 0;
+	int ret = 0;
 
 retry:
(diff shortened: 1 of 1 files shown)


neocturne:
That code comes from upstream ebtables. It was introduced in http://git.netfilter.org/ebtables/commit/libebtc.c?id=6a826591878db3fa9e2a94b87a3d5edd8e0fc442. As far as I can tell, the change of behavior was intentional, they just forgot to update the comment.

You can propose a fix for the comment upstream if you care enough, but ebtables is pretty much dead at this point, with not commits in over a year - nftables is a much cleaner and more powerful solution for bridge filtering. Gluon is going to drop ebtables in favor of nftables as well.

PR #271, opened 2024-01-18
wgpeerselector: fix undefined variable peer

maurerle:
This fixes a minor bug, which happens, if the key has been invalidated by the server and the connection to the gateway is lost due to the gateway closing the wireguard tunnel.

A trace is logged on the node due to the undefined variable coming from a copy-paste mistake from a few lines above.

This is fixed by using the `connected_peer`.

### DIFF
--- a/net/wgpeerselector/files/usr/bin/wgpeerselector
+++ b/net/wgpeerselector/files/usr/bin/wgpeerselector
@@ -380,7 +380,7 @@ function WGPeerSelector:main()
 
 			if self:try_connect_to_peer(peer, timeout) then
 				connected_peer = peer
-				log(syslog.LOG_INFO, 'Connection established with '..peer.name..'.')
+				log(syslog.LOG_INFO, 'Connection established with '..connected_peer.name..'.')
 			end
 
 		elseif state == 'established' then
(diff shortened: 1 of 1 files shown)


net/wgpeerselector/files/usr/bin/wgpeerselector
```diff
@@ -389,7 +389,7 @@ function WGPeerSelector:main()
 			if not connected_peer:has_recent_handshake() then
 				connected_peer:uninstall_from_kernel()
 				connected_peer = nil
-				log(syslog.LOG_INFO, 'Connection to '..peer.name..' lost.')
+				log(syslog.LOG_INFO, 'Connection to '..connected_peer.name..' lost.')
```
grische, line 392: Might be worth changing both 383 and 393 to use `connected_peer` instead?

grische (approved):
LGTM

net/wgpeerselector/files/usr/bin/wgpeerselector
blocktrron, line 392: You set this to nil in the previous line?

skorpy2009, line 392: Why are you using a variable which is set to nil here?

maurerle: At least, it is declared ;)

PR #274, opened 2024-03-08
autoupdater: update download progress display less frequently

### LINKED ISSUE #273: autoupdater: extremely slow image download on HiveAP
I'm currently using autoupdater to update the firmware on an Aerohive HiveAP 330. Which takes ages.
So I've used that time to have a look what's holding it up.
The HiveAP 330 uses a 9600bps console output, so any output takes a long time.

The while loop in recv_image_cb outputs the progress in every iteration:
https://github.com/freifunk-gluon/packages/blob/3d08b0fee8dc5d96d8bcdb985fad1d5564de4022/admin/autoupdater/src/autoupdater.c#L252-L277

Maybe it would be beneficial if the downloading progress would only get shown, like, every 128th iteration or so?

On a sidenote, it also looks to me like the SHA256 checksum is calculated for every iteration as well? Is this necessary? Wouldn't it be sufficient to calculate it once the download is completed?
https://github.com/freifunk-gluon/packages/blob/3d08b0fee8dc5d96d8bcdb985fad1d5564de4022/admin/autoupdater/src/autoupdater.c#L275

Okay, it finally finished downloading, took about 20 minutes for the 44MB image...

neocturne:
Updating the "XXXXX / XXXXX KiB" display for every run of recv_image_cb() is a significant bottleneck on slow consoles. This was reported for a device using a 9600 Baud serial console, but the slowdown is noticeable even on a 115200 Baud console, especially when the network is fast.

Replace the output with a "XX.X / XX.X MiB" display and only update it every 0.1 MiB to fix the issue.

Closes #273

### DIFF
--- a/admin/autoupdater/src/autoupdater.c
+++ b/admin/autoupdater/src/autoupdater.c
@@ -72,6 +72,7 @@ struct recv_manifest_ctx {
 struct recv_image_ctx {
 	int fd;
 	ecdsa_sha256_context_t hash_ctx;
+	ssize_t downloaded;
 };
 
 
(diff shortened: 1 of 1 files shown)


rotanid:
@neocturne @kpanic23 what about this, has this been tested already?

neocturne:
> @neocturne @kpanic23 what about this, has this been tested already?

I have tested it on one of my devices with a 115200 Baud UART, but it would be great to hear how well it works at 9600 Baud.

Djfe:
I have an Aerohive ap121 here and could test this patch

rotanid:
@Djfe that statement might be true for many, including me, but it doesn't matter at all until someone really does it :-P

neocturne:
@T-X Would you still say the solution is good enough with the numbers I provided?

T-X:
So for an 8 MiB image at a 9600 baud console (worst case) output would need `(8*10*50)/(9600/10) = 4.2` seconds. Typically, at 115200 baud, more like 0.03 seconds. Sounds negligible for now in comparison to the total update time. And I like the simplicity/readability of this change, so sounds good to me.

rotanid:
> 44MiB are probably a typo, and the image is actually 4 or 4.4MiB?

no typo, see https://github.com/freifunk-gluon/packages/issues/273
also, our own images for HiveAP are also ~16MB for v2023.1.x based builds and ~44 for v2021.1.x based builds.

PR #282, opened 2026-08-10
autoupdater: add ssl context if supported

maurerle:
This fixes the issue that the autoupdater does not support https urls.

Tested on AP-325 with `autoupdater -f https://....`

### DIFF
--- a/admin/autoupdater/src/uclient.c
+++ b/admin/autoupdater/src/uclient.c
@@ -9,6 +9,8 @@
 
 #include <limits.h>
 #include <stdio.h>
+#include <string.h>
+#include <glob.h>
 
 
 #define TIMEOUT_MSEC 300000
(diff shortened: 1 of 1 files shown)


admin/autoupdater/src/uclient.c
```diff
@@ -151,11 +159,30 @@ int get_url(const char *url, void (*read_cb)(struct uclient *cl), void *cb_data,
...
+		unsigned int i;
+
+		if (glob("/etc/ssl/certs/*.crt", 0, NULL, &gl) == 0) {
+			for (i = 0; i < gl.gl_pathc; i++)
+				ssl_ops->context_add_ca_crt_file(ssl_ctx, gl.gl_pathv[i]);
+			globfree(&gl);
+		}
+
+		uclient_http_set_ssl_ctx(cl, ssl_ops, ssl_ctx, true);
+	} else if (!strncmp(url, "https://", 8)) {
```
T-X:
```suggestion
	} else if (!strncmp(url, "https://", strlen("https://"))) {
```

blocktrron (commented):
Did not run-test but looks sensible to me.

T-X (changes requested):

PR #283, opened 2026-08-16
lua-geolocate: update URL to use https

SergejDurov1987:
As we are using https elsewhere already on the nodes, let's also switch this URL to use https.

### DIFF
--- a/libs/lua-geolocate/src/geolocate.lua
+++ b/libs/lua-geolocate/src/geolocate.lua
@@ -66,7 +66,7 @@ local function locate(blacklist)
 
 	assert(#found_bssids >= 12, 'insufficient BSSIDs found')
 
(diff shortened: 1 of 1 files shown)


rotanid:
@SergejDurov1987 are you sure every node running this also has all the necessary packages for making that HTTPS connection?

SergejDurov1987:
> @SergejDurov1987 are you sure every node running this also has all the necessary packages for making that HTTPS connection?

Good call! I've implemented a helper function that checks whether an SSL/TLS library is available on the node, and then decides whether to use SSL or not based on that.

T-X:
Is that service actually still working? The website, https://openwifi.su/, looks different now. @SergejDurov1987, have you tested this change?

PR #285, opened 2026-08-29
autoupdater: save manifest to /tmp/manifest.txt

T-X:
So far the Gluon autoupdater manifest file is downloaded and evaluated line-wise, on the fly within the autoupdater. This makes it difficult to examine or preserve a downloaded manifest file used in specific autoupdater attempt.

Similar to the /tmp/firmware.bin download let's also keep a copy of the manifest file in /tmp/manifest.txt. The manifest.txt file size is limited to 512KB. This file can then be examined after parsing errors or when using "--no-action". Or can be preserved over updates via a a script in /usr/lib/autoupdater/upgrade.d later.


admin/autoupdater/src/autoupdater.c
```diff
@@ -191,7 +194,7 @@ static float get_probability(time_t date, float priority, bool fallback) {
 static void recv_manifest_cb(struct uclient *cl) {
 	struct recv_manifest_ctx *ctx = uclient_get_custom(cl);
 	char *newline;
-	int len;
+	int total_len = 0, len;
```
maurerle, line 197: the `recv_manifest_cb` might be called multiple times (as a slow or adversarial server sending data in tiny trickles would bypass it), in which case the total_len is not reset.
We should move the total_len declaration to the `recv_manifest_ctx` struct

admin/autoupdater/src/autoupdater.c
```diff
@@ -285,6 +302,11 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
 	char manifest_url[strlen(mirror) + strlen(s->branch) + 11];
 	sprintf(manifest_url, "%s/%s.manifest", mirror, s->branch);
 
+	manifest_ctx.fd = open(manifest_path, O_WRONLY|O_CREAT, 0600);
```
maurerle, line 305:
```suggestion
	manifest_ctx.fd = open(manifest_path, O_WRONLY|O_CREAT|O_TRUNC, 0600);
```
we should add `O_TRUNC` to fix an existing file containing garbage.

maurerle (changes requested):
Thats a good idea!
I left two improvements to address before merge

admin/autoupdater/src/autoupdater.c
T-X: I'm having some trouble finding the place in the (of course undocumented...) uclient library where it supposedly calls the `recv_manifest_cb` / the callback function multiple times. Do you have some pointers? (My initial assumption was that it wouldn't and that that's why the while-loop in recv_manifest_cb exists.)

admin/autoupdater/src/autoupdater.c
T-X: Good point. I think we might have a similar problem with writing the firmware to /tmp/firmware.bin. (Though at least this open()/write() handling with multiple calls to the autoupdater is currently not appending, but overwriting from byte 0. So multiple calls in most cases would result in a working firmware file, I guess.)

I would just replace the open() calls for manifest and firmware files with creat() calls then?

T-X: Ah, no, we don't have a problem for a firmware.bin. The manifest specifies and limits the firmware.bin download size. And a /tmp/firmware.bin is deleted/unlinked if it's not the correct size or checksum.

admin/autoupdater/src/autoupdater.c
T-X: Nevermind, I think I see the loop, probably the one in ustream_fd_read_pending() in libubox:

```
ustream_fd_read_pending()
    do {
        ustream_fill_read()
        -> s->notify_read()
            -> uclient_notify_read()
                 -> __uclient_notify_read()
                     -> uclient_backend_read_notify()
                         -> read_notify()
                             -> __uclient_read_notify()
                                 -> data_read()
                                     -> recv_manifest_cb()
    } while (1);
```

Will move total_len into the recv_manifest_ctx as suggested, thanks!

T-X:
Changelog v2:
* replaced `manifest_ctx.fd = open(manifest_path, O_WRONLY|O_CREAT, 0600);`  with `manifest_ctx.fd = creat(manifest_path, 0600);` to avoid residual bytes from an earlier attempt
* moved `total_len` into `struct recv_manifest_ctx`, to ensure a proper continuation/counting if libubox ustream calls the callback for the same get_url() call multiple times

PR #286, opened 2026-08-29
autoupdater: allow overwriting mirrors via manifest file

T-X:
Allow overwriting the URLs to download an image file from within a signed manifest file. The new attribute is "MIRROR=" and can be specified multiple times. These mirrors are tried in their specified order, similar to how the CLI provided mirror overwrites are tried in order.

---

The primary security promise lies in a signed manifest file. And its n-of-m signatures approach to facilitate a democratic security architecture.

Technically the autoupdater server side, including its DNS can currently have a separate governance. And has the power to overrule or disrupt any democratic decisions from the signers, challenging their souvereignity. In the worst case a single person running the autoupdater HTTP server and/or its DNS entries can overrule the signers collective decision. Also due to a misconfiguration or hardware failure an autoupdater server might accidentally become unreachable.

This patch is in preparation to later allow alternative sources for the autoupdater images. For instance via a pushed manifest file. Or via node-to-node autoupdates.


admin/autoupdater/src/autoupdater.c
```diff
@@ -272,29 +272,24 @@ static void recv_image_cb(struct uclient *cl) {
...
 
 	/* Download manifest */
 	ecdsa_sha256_init(&m->hash_ctx);
 	int err_code = get_url(manifest_url, recv_manifest_cb, &manifest_ctx, -1, s->old_version);
 	if (err_code != 0) {
 		fprintf(stderr, "autoupdater: warning: error downloading manifest: %s\n", uclient_get_errmsg(err_code));
-		interrupted = uclient_interrupted_signal(err_code);
-		goto out;
+		*interrupted = uclient_interrupted_signal(err_code);
+		return -EINVAL;
```
maurerle, line 292: This requires importing `#include <errno.h>` to build

T-X:
Changelog v2:
* adding `#include <errno.h>`

T-X:
One more thought/question I'd have is priorities. Right now manifest embedded mirrors would have the highest preference and UCI mirrors the lowest. So: embedded mirrors > CLI mirrors > UCI mirrors. (And if a mirror set of higher preference exists then mirror sets of lower preference are ignored.)

Is that what users would want/expect? Or should it be: CLI mirrors > embedded mirrors > UCI mirrors? Or maybe, alternatively a CLI option like "--ignore-embedded-mirrors" could be useful for CLI users?

rotanid:
- `run_dir(download_d_dir)` moved inside `download_and_verify_image()`, so with embedded mirrors it fires once per mirror attempt instead of once per cycle. Need to confirm `download.d/*` scripts are idempotent, or hoist it back to once-per-run.
- firmware_path is reopened with O_WRONLY|O_CREAT (no O_TRUNC/unlink) between mirror attempts. A shorter successor can't overwrite leftover tail bytes from a longer failed predecessor. hash verification catches it, so not a correctness hole, but O_TRUNC would make it tidy.

On the precedence question: since it's a community-values decision, maybe make it a UCI option rather than hardcoding. Default to today's behavior (UCI wins), with an embedded opt-in for communities that want signer-chosen mirrors to take precedence. Ships via site config, no silent behavioral shift.

(and this one should be merged before #289 - as the latter anyways kind of depends on 286 and needs rebase afterwards)

admin/autoupdater/src/autoupdater.c
```diff
@@ -394,11 +392,52 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
...
+	if (!m->n_mirrors)
+		return download_and_verify_image(mirror, s, m, interrupted);
+
+	/* mirrors in manifest files replace mirrors from CLI or UCI */
+	for (size_t i = 0; i < m->n_mirrors; i++) {
+		mirror = m->mirrors[i];
+		if (!mirror)
+			continue;
+
+		ret = download_and_verify_image(mirror, s, m, interrupted);
```
maurerle, line 415: When a download is interrupted by the user pressing `Ctrl-C`, `download_and_verify_image()` sets `*interrupted = SIGINT` and returns `-EINVAL`. However,  `download_and_verify_images()` only checks `if (!ret)` and does not inspect `*interrupted`, causing the loop to immediately attempt the next mirror in the list.

If a subsequent mirror succeeds quickly, the autoupdater proceeds to invoke `sysupgrade` even though the user explicitly requested to abort.

admin/autoupdater/src/autoupdater.c
```diff
@@ -308,48 +304,50 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
...
 	}
 
-	/**** Download and verify image file *********************************/
+	return 0;
+}
+
+static int download_and_verify_image(const char *mirror, struct settings *s, struct manifest *m,
+				     int *interrupted) {
 	/* Begin download of the image */
 	run_dir(download_d_dir);
```
maurerle, line 344: Call `run_dir(download_d_dir)` once before entering the mirror retry loop rather than inside the per-mirror download function.

abort.d is only called once

admin/autoupdater/src/autoupdater.c
```diff
@@ -308,48 +304,50 @@ static bool autoupdate(const char *mirror, struct settings *s, int lock_fd) {
...
 
 	struct recv_image_ctx image_ctx = { .downloaded = -1 };
 	image_ctx.fd = open(firmware_path, O_WRONLY|O_CREAT, 0600);
```
maurerle, line 347: As this is now called in a loop, we need to either ensure: `O_WRONLY | O_CREAT | O_TRUNC`

Or run `unlink(firmware_path);` between attempts.

maurerle (changes requested):

maurerle:
@T-X I think that if we go that route, we need to have something like: `--ignore-embedded-mirrors` to force a behavior on CLI, but typically expect to prioritize embedded mirror so `embedded mirrors > CLI mirrors > UCI mirrors`.

But I am also a little bit unsure if this is really needed on top of TLS.
Of course, this is to handle developer intent to validate where the update is coming from primarily.

PR #289, opened 2026-09-04
autoupdater: allow using stdin for a manifest CLI mirror

T-X:
Allow reading a manifest file from stdin instead of an HTTP autoupdater mirror when using the CLI by using "-" as a mirror CLI argument. For instance like:

  $ cat /root/experimental.manifest | autoupdater -

This can be useful for development and debugging for one thing. For another the idea is to later use it for a push-induced autoupdate mechanism.


rotanid (commented):
Nice refactor overall. One real problem though:

After parsing a stdin manifest, the image download still constructs the URL as `%s/%s` with `mirror`, so for `-` it tries to fetch `"-/<image-filename>"` via `get_url`, which is HTTP-only — that fails at connect time. Net effect: `cat foo.manifest | autoupdater -` only "works" when the manifest says nothing-new-to-do; the moment it actually finds newer firmware (and the proceed checks pass) it dies with a misleading `error downloading image: Connection failed`. And `-n` doesn't dodge it either, since it still downloads the image, just skips flashing.

Probably worth deciding intentionally what `-` means for the image step: reject it with a clear message when an update would proceed, treat `-` as manifest-validate-only, or fold the image delivery into the companion gluon PR #3824 and skip the HTTP fetch here. Right now the failure is neither guarded nor clearly diagnosed.

Two smaller things:
- In `recv_manifest_cb` any `len <= 0` from `read_cb` is treated as clean EOF, so `read_stdin_manifest_cb` returning `-EINVAL` on `ferror(stdin)` is indistinguishable from normal end-of-input. With stdin a broken pipe mid-read is realistic, so maybe distinguish EOF (0) from error (<0) and route it to `err_code` so the existing error handler prints something honest instead of a downstream "missing mandatory fields".
- Cosmetic: prints `Retrieving manifest from - ...`; `stdin` would read nicer.

rotanid (commented):
duplicate — please disregard

T-X:
> After parsing a stdin manifest, the image download still constructs the URL as `%s/%s` with `mirror`, so for `-` it tries to fetch `"-/<image-filename>"` via `get_url`, which is HTTP-only — that fails at connect time. Net effect: `cat foo.manifest | autoupdater -` only "works" when the manifest says nothing-new-to-do; the moment it actually finds newer firmware (and the proceed checks pass) it dies with a misleading `error downloading image: Connection failed`. And `-n` doesn't dodge it either, since it still downloads the image, just skips flashing.

That's where maybe https://github.com/freifunk-gluon/packages/pull/286 would be useful, to use in combination with this PR. That's how I had tested it so far. But other than that, good point.

Right now, with #286, we iterate as follows:

```
for/try each CLI/UCI mirror m:
-> get manifest via m
a) if manifest embedded mirrors:
    -> for/try each manifest embedded mirror M:
        -> get image via M
b) if no manifest embedded mirror:
    -> get image via m
```

I'm wondering if it would make sense, if no manifest embedded mirrors are specified to once more iterate over all CLI/UCI mirrors in the image download step:


```
for/try each CLI/UCI mirror m:
-> get manifest via m
a) if manifest embedded mirrors:
    -> for/try each manifest embedded mirror M:
        -> get image via M
b) if no manifest embedded mirror:
    -> for/try each CLI/UCI mirror M:
        -> get image via M
```

rotanid:
ok, sounds reasonable, but also the iteration has to exclude the "-" then. and still, having nice error messages would be preferrable.

PR #291, opened 2026-09-05
libolsrdhelper: init

mkg20001:


utils/libolsrdhelper/src/neigh.c
```diff
@@ -0,0 +1,192 @@
+/* SPDX-FileCopyrightText: 2021-2023 Maciej Krüger <maciej@xeredo.it> */
```
mkg20001, line 1: TODO: rewrite away from mac as neighbour to ip as neighbour

